PR #43, opened 2015-04-02
Drop reference to moved project

DavidPrevot:
Hi Evert,

Please note that you may wish to simply drop the copyright notice from the [generic license](http://sabre.io/license/), or at least update the last year there too.


evert:
Thanks! didn't realize this was outdated. Will update the other year as well.

PR #44, opened 2015-05-29
Prevent eternal loop by ensuring that no libxml errors with level [LIBXML_ERR_FATAL]  have occurred

DaanBiesterbos:
Hi, 
I was facing some instability issues regarding invalid XML documents. I would expect an exception but instead the Reader will (in some cases) attempt to keep parsing the same node eternally. I fixed this by adding a validation step.  The XMLReader::isValid may not be backward compatible so I made the validation step optional. Altough I did not validate this thouroughly it seems like this method will only work when the something like a XSD, DTD etc is in place.  Either way. The feature is optional, so this change should not be problematic. 

Update:
The validation did not fix all scenario's. The new solution does. Basically we now check if libxml has fatal errors on each iteration. I did not implement this as a optional feature since it ensures that your application will not crash when invalid XML is provided.

### DIFF
--- a/lib/Reader.php
+++ b/lib/Reader.php
@@ -55,6 +55,7 @@ class Reader extends XMLReader {
      * This function will also disable the standard libxml error handler (which
      * usually just results in PHP errors), and throw exceptions instead.
      *
+     * @throws LibXMLException
      * @return array
      */
     function parse() {
@@ -76,10 +77,9 @@ class Reader extends XMLReader {
 
         if ($errors) {
             throw new LibXMLException($errors);
-        } else {
-            return $result;
         }
 
+        return $result;
     }
 
     /**
@@ -93,16 +93,14 @@ class Reader extends XMLReader {
      * If the $elementMap argument is specified, the existing elementMap will
      * be overridden while parsing the tree, and restored after this process.
      *
+     * @throws ParseException
      * @param array $elementMap
      * @return array|string
      */
     function parseInnerTree(array $elementMap = null) {
 
-        $previousDepth = $this->depth;
-
         $text = null;
         $elements = [];
-        $attributes = [];
 
         if ($this->nodeType === self::ELEMENT && $this->isEmptyElement) {
             // Easy!
@@ -255,7 +253,7 @@ class Reader extends XMLReader {
      * short keys. If they are defined on a different namespace, the attribute
      * name will be retured in clark-notation.
      *
-     * @return void
+     * @return array
      */
     function parseAttributes() {
 
@@ -281,22 +279,5 @@ class Reader extends XMLReader {
         return $attributes;
 
     }
-
-    /**
-     * @return bool
-     */
-    function isValidationEnabled()
-    {
-        return $this->validationEnabled;
-    }
-
-    /**
-     * @param bool $enable
-     */
-    function setValidationEnabled($enable = true)
-    {
-        $this->validationEnabled = $enable;
-    }
-
 
 }
--- a/lib/Reader.php
+++ b/lib/Reader.php
@@ -21,9 +21,6 @@ class Reader extends XMLReader {
 
     use ContextStackTrait;
 
-    /** @var bool  */
-    protected $validationEnabled = false;
-
     /**
      * Returns the current nodename in clark-notation.
      *
--- a/lib/Reader.php
+++ b/lib/Reader.php
@@ -122,19 +122,12 @@ class Reader extends XMLReader {
         // https://bugs.php.net/bug.php?id=64230
         if (!@$this->read()) return false;
 
-        while (true) {
+        while ($this->canContinue()) {
 
             switch ($this->nodeType) {
                 case self::ELEMENT :
-                    // I have encountered a situation where the document was deemed valid while there was a syntax issue with a given end tag. An eternal loop was the result.
-                    // XMLReader will never validate the entire tree. Exclusively the node that is processed. Each node should be individually validated to make sure that the parser ends processing unexpected (syntax)
-                    // errors occur. This validation should be  optional and disabled by default for backward compatibility. I'm did not test every
-                    // use case I could think of, but it seems like the validation will always fail without the proper schema (or DTD) in place.
-                    if(!$this->isValidationEnabled() or $this->isValid()) {
-                        $elements[] = $this->parseCurrentElement();
-                        break;
-                    }
-                    throw new ParseException("Unable to parse invalid '{$this->localName}' node. The XML document is not valid.");
+                    $elements[] = $this->parseCurrentElement();
+                    break;
                 case self::TEXT :
                 case self::CDATA :
                     $text .= $this->value;
@@ -147,8 +140,11 @@ class Reader extends XMLReader {
                 case self::NONE :
                     throw new ParseException('We hit the end of the document prematurely. This likely means that some parser "eats" too many elements. Do not attempt to continue parsing.');
                 default :
-                    // Advance to the next element
-                    $this->read();
+                    // This prevents an eternal loop in case of a missing open tag.
+                    if(!$this->read()) {
+                        throw new ParseException("Unable to parse invalid '{$this->localName}' node. The XML document is not valid.");
+                        break 2; // Break out of invalid element
+                    }
                     break;
             }
 
@@ -159,6 +155,31 @@ class Reader extends XMLReader {
         }
         return ($elements ? $elements : $text);
 
+    }
+
+    /**
+     * Check internal (libxml) errors. Throw an exception if needed.
+     * We need to do this for every node or won't be able to recognize malformed XML.
+     * This would lead to an eternal loop.
+     *
+     * @throws ParseException
+     * @return bool
+     */
+    protected function canContinue()
+    {
+        // Check for errors. Normally there will be no error, so this should not have a big impact on performance.
+        $errors = libxml_get_errors();
+        if(count($errors)) {
+            /** @var \LibXMLError $error */
+            foreach($errors as $e) {
+                if ($e->level === LIBXML_ERR_FATAL){
+                    libxml_clear_errors();
+                    throw new LibXMLException($errors);
+                }
+            }
+        }
+
+        return true;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


Hywan:
Hello :-),

First, can you give us an example of your XML in order to reproduce your issue please?
Second, do you have any idea of the impact on performances?
Third, is it possible to validate the whole document once instead of each node?

evert:
Damn... lost my work in progress comment.

@Hywan One of the key concepts of the XMLReader/XMLWriter is that they can operate on streams. So if validation is possible for the whole document, I don't think it would be desired. I'm also not super concerned about speed if it's an optional feature.

@DaanBiesterbos 

A few points:

* @Hywan is right that this change needs a test. A unittest that checks both the valid and invalid case.
* You can automatically fix all CS-related issues by running : `./bin/sabre-cs-fixer fix .`
* The docblocks should be expanded a bit to explain exactly what the feature does, and how to use it.
* Don't return `$this` from the setter.

evert:
Lastly... is it possible to emit a more useful exception? It would be great if we can tell the user *why* their document isn't valid... i assume that XMLReader has a way to access this information?

DaanBiesterbos:
@Hywan   @evert    Hi guys. I will prepare something to reproduce the problem and improve the commit as described above. As evert pointed out the isValid method will always only check the current node. But there is also a validation to check if the XML is valid according to the schema. This validation does work for the entire tree. This validation detects nodes that violate the schema but does completely ignore syntax errors. 
I'll come back on this later. One more thing though. 

I executed the the ./bin/sabre-cs-fixer fix command. But a LogicException was thrown in ./vendor/symfony/finder/Symfony/Component/Finder/Finder.php on line 698.  Message: "You must call one of in() or append() methods before iterating over a Finder."
$config->getDir() returns null in "/home/daan/Projecten/SabreXmlReader/sabre-xml/vendor/fabpot/php-cs-fixer/Symfony/CS/Fixer.php" on line 137.

It seems like adding '$out->setDir(__DIR__ . '/../lib');' to  "/home/daan/Projecten/SabreXmlReader/sabre-xml/vendor/sabre/cs/bin/php-cs-fixer-config.php" solved the problem locally. I can now execute the command. Not sure what is causing this though. I tried downgrading to symfony 2.6.7, but that did not solve it.

evert:
Did you do `sabre-cs-fixer fix .` ? the last `.` is important =)

DaanBiesterbos:
@evert  Ah. You are right. Missed the dot. :-)

DaanBiesterbos:
To reproduce the issue you can run the unit test and omit the solution :) All scenario's that should fail because of missing tags or malformed XML will lead to an eternal loop.

DaanBiesterbos:
@evert   Hi Evert. How do you feel about the provided solution? Were you able to reproduce the problem?

evert:
Hi Daan,

The reason I have not really gotten to it yet, is because this is a relatively large diff, for a relatively small change. There's a lot of unrelated changes in the source, and things could probably get simplified a bit.

Will make a few comments in-line.

evert:
The biggest thing I would want to ask is.. how small can you make this unittest? Right now there's a ton of assertions, and a lot of other stuff. Could you create a very minimal schema?

lib/Reader.php
```diff
@@ -38,7 +38,6 @@ function getClark() {
         }
 
         return '{' . $this->namespaceURI . '}' . $this->localName;
-
```
evert: Unrelated change, revert this.

lib/Reader.php
```diff
@@ -73,10 +73,9 @@ function parse() {
 
         if ($errors) {
             throw new LibXMLException($errors);
-        } else {
-            return $result;
         }
 
+        return $result;
```
evert: Not bad, but unrelated change... please revert this to simplify the patch.

lib/Reader.php
```diff
@@ -113,13 +110,14 @@ function parseInnerTree(array $elementMap = null) {
         }
 
 
-        // Really sorry about the silence operator, seems like I have no
-        // choice. See:
+        // Really sorry about the silence operator, seems like I have no choice. See:
```
evert: Unrelated change, please revert

lib/Reader.php
```diff
@@ -147,8 +144,33 @@ function parseInnerTree(array $elementMap = null) {
...
+     *
+     * @throws ParseException
+     * @return bool
+     */
+    protected function canContinue()
+    {
+        // Check for errors. Normally there will be no error, so this should not have a big impact on performance.
+        $errors = libxml_get_errors();
+        if (count($errors)) {
+            /** @var \LibXMLError $error */
```
evert: `$error` is not defined, should be `$e`.

tests/Sabre/Xml/ParserTest.php
```diff
@@ -0,0 +1,587 @@
...
+namespace Sabre\Xml;
+
+/**
+ * Test the XML reader.
+ * -Test with valid document
+ * -Test XML schema validation
+ * -Test invalid documents (missing start and end tag, malformed start and end tag and malformed quotes)
+ *
+ *
+ * PHP version 5
```
evert: docblocks should match sabre/dav's docblocks, or be removed (we don't require them in unittests).

tests/Sabre/Xml/ParserTest.php
```diff
@@ -0,0 +1,587 @@
...
+ *
+ * Class ParserTest
+ * @author     Daan Biesterbos <daanbiesterbos@gmail.com>
+ * @license http://sabre.io/license/ Modified BSD License
+ * @since     1.0.0
+ */
+class ParserTest extends \PHPUnit_Framework_TestCase
```
evert: Rename to something like 'ReaderSchemaValidationTest`.

tests/Sabre/Xml/ParserTest.php
```diff
@@ -0,0 +1,587 @@
...
+        if (function_exists('register_tick_function')) {
+            // Register tick function
+            register_tick_function(function() {
+                throw new \LogicException("Test failed. The reader seems to be trapped in a eternal loop. Failed to recognize the missing start tag...");
+            });
+        }
+
+
+        // Parse elements, use tick counter to break out of eternal loop and throw an exception
+        declare (ticks = 10000000); // Don't use low values, apparently phpunit is also using ticks and will collide with low values)
```
evert: I would suggest to get rid of this altogether. It's good that the unittest can report that the script has entered an eternal loop, but the unittests simply not completing is a good enough indicator for me. This is what we do in sabre/vobject for a lot of stuff. Travis will just time out for these tests and a developer can just abort the script... No biggy and avoids stuf like ticks.

tests/Sabre/Xml/ParserTest.php
```diff
@@ -0,0 +1,587 @@
...
+    function testParserShouldSucceed()
+    {
+        $reader = new Reader();
+        $reader->XML($this->xmlValidDocument);
+        $result = $reader->parse();
+        $this->assertNotEmpty($result);
+        $result = $result['value'];
+
+        // Validate store name
+        $this->assertEquals('Awesome Shop Inc.', $result[0]['value'], "The store name is not valid.");
```
evert: Could the entire `$result` array just be checked for as a whole. This will make this a _lot_ clearer I think.

tests/Sabre/Xml/ParserTest.php
DaanBiesterbos: Hi Evert, 
My idea was to write a test to parse various valid and invalid formats. I also used this for development. I could refactor this and make it more compact. I'm not sure about the name though. I removed the schema validation entirely. The goal was to get rid of the eternal loops, and this test assures they do not occur.  That solution seems to be pretty solid. (and a lot faster than schema validation! :+1: )  What do you think about "ReadInvalidDocumentTest" - as most of the tests are all about testing invalid xml anyway.  A compact version of the test on top (valid format) could be moved to another test case. (or removed entirely if you feel like testing valid formats is covered by the other tests?)

tests/Sabre/Xml/ParserTest.php
DaanBiesterbos: Do you mean just the ticks part?   These tests were very useful during development.  I tried some other things and this tests quickly showed me which invalid formats would cause an eternal loop, and which did not. I have to admit that although the solution is easy it was a bit tricky to find a solution that worked under all circumstances. This tests are the main reason if not the only one to be confident about the solution. But if are sure you don't need the tests than I don't mind to remove them. Your call. :-)

tests/schema/Store.xsd
```diff
@@ -0,0 +1,31 @@
...
+    <!--+====================================================================================================+-->
+    <!--+ Elements                                                                                           +-->
+    <!--+====================================================================================================+-->
+    <xs:element name="name" type="xs:string" />
+    <xs:element name="customers" type="person:CustomerCollection" />
+    <xs:element name="store" type="store:StoreType" />
+
+
+
+</xs:schema>
```
DaanBiesterbos: I guess the schema's can be removed, unless you want to keep them for a future test. ;-)

tests/Sabre/Xml/ParserTest.php
evert: I wouldn't say: get rid of the tests, but rather just get rid of the code that can abort the test if it took too long. I'm ok with the unittests getting stuck in a loop if the code is incorrect, as long as they complete when the test _is_ good.

I just want to simplify.

tests/Sabre/Xml/ParserTest.php
evert: That sounds like a good name to me too =) Just wanted it to be a bit more descriptive.

tests/schema/Store.xsd
evert: It would be nice to do a schema-related unittest, but in that case I would want to create an "as small as possible" schema.a Unittests should all be about the testing the most isolated possible cases.

evert:
Hi @DaanBiesterbos ,

Didn't hear back in a while, and then another related issue popped up in #45.
In response to that, I created a new patch that's similar to yours, but much simpler.

Tagging this today as 1.1.0.

PR #51, opened 2015-08-10
Disable entity loader

LukasReschke:
Otherwise it is possible to perform requests to other domains, gaining access to local file resources seems not possible.

Following XML will perform a request to http://owncloud.org when the loader is enabled which is especially of relevance due to https://bugs.php.net/bug.php?id=64938.
```xml
<?xml version="1.0" encoding="utf-8" standalone='no'?>
<!DOCTYPE books [
<!ENTITY % e1fe9 SYSTEM "http://owncloud.org">%e1fe9; ]>
<books
    xmlns="http://example.org/books">
    <book>
        <title>Snow Crash</title>
        <author>Neil Stephenson</author>
    </book>
    <book>
        <title>Dune</title>
        <author>Frank Herbert</author>
    </book>
</books>
```

@evert THX

### DIFF
--- a/lib/Reader.php
+++ b/lib/Reader.php
@@ -56,6 +56,7 @@ class Reader extends XMLReader {
      */
     function parse() {
 
+        $previousEntityState = libxml_disable_entity_loader(true);
         $previousSetting = libxml_use_internal_errors(true);
 
         // Really sorry about the silence operator, seems like I have no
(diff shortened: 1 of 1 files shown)


LukasReschke:
Just to clarify: Very low severity. No CVE or advisory required here from my PoV. Just some best practice enhancements.

evert:
I agree on severity. The external entity loader is disabled by default for some time now, so it would only affect those really few people that actually use it.

https://github.com/php/php-src/commit/8e76d0404b7f664ee6719fd98f0483f0ac4669d6

staabm:
Side note: it affects everyone not using php7...?

LukasReschke:
> The external entity loader is disabled by default for some time now, so it would only affect those really few people that actually use it.

Tell this all the awesome distributions which do not update their PHP. Because backporting stuff is obviously bad *sigh*

/me is gonna be having a big party once PHP 5.x is finally dead for the good… :speak_no_evil:

PR #57, opened 2015-10-31
Remove unused variables

jakobsack:
This pull request removes two variables that are initialized but never used.


evert:
Excellent. Thank you!

PR #61, opened 2015-12-02
External object serializers.

evert:
cc: @staabm. What do you think?

### DIFF
--- a/lib/ContextStackTrait.php
+++ b/lib/ContextStackTrait.php
@@ -58,6 +58,26 @@ trait ContextStackTrait {
     public $namespaceMap = [];
 
     /**
+     * This is a list of custom serializers for specific classes.
+     *
+     * The writer may use this if you attempt to serialize an object with a
+     * class that does not implement XmlSerializable.
+     *
+     * Instead it will look at this classmap to see if there is a custom
+     * serializer here. This is useful if you don't want your value objects
+     * to be responsible for serializing themselves.
+     *
+     * The keys in this classmap need to be fully qualified PHP class names,
+     * the values must be callbacks. The callbacks take two arguments. The
+     * writer class, and the value that must be written.
+     *
+     * function (Writer $writer, object $value)
+     *
+     * @var array
+     */
+    public $classMap = [];
+
+    /**
      * Backups of previous contexts.
      *
      * @var array
@@ -78,7 +98,8 @@ trait ContextStackTrait {
         $this->contextStack[] = [
             $this->elementMap,
             $this->contextUri,
-            $this->namespaceMap
+            $this->namespaceMap,
+            $this->classMap
         ];
 
     }
@@ -93,7 +114,8 @@ trait ContextStackTrait {
         list(
             $this->elementMap,
             $this->contextUri,
-            $this->namespaceMap
+            $this->namespaceMap,
+            $this->classMap
         ) = array_pop($this->contextStack);
 
     }
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -39,6 +39,26 @@ class Service {
     public $namespaceMap = [];
 
     /**
+     * This is a list of custom serializers for specific classes.
+     *
+     * The writer may use this if you attempt to serialize an object with a
+     * class that does not implement XmlSerializable.
+     *
+     * Instead it will look at this classmap to see if there is a custom
+     * serializer here. This is useful if you don't want your value objects
+     * to be responsible for serializing themselves.
+     *
+     * The keys in this classmap need to be fully qualified PHP class names,
+     * the values must be callbacks. The callbacks take two arguments. The
+     * writer class, and the value that must be written.
+     *
+     * function (Writer $writer, object $value)
+     *
+     * @var array
+     */
+    public $classMap = [];
+
+    /**
      * Returns a fresh XML Reader
      *
      * @return Reader
@@ -60,6 +80,7 @@ class Service {
 
         $w = new Writer();
         $w->namespaceMap = $this->namespaceMap;
+        $w->classMap = $this->classMap;
         return $w;
 
     }
--- a/lib/Writer.php
+++ b/lib/Writer.php
@@ -100,8 +100,14 @@ class Writer extends XMLWriter {
 
         if (is_scalar($value)) {
             $this->text($value);
-        } elseif ($value instanceof XmlSerializable) {
+        } elseif (is_object($value) && $value instanceof XmlSerializable) {
+
             $value->xmlSerialize($this);
+
+        } elseif (is_object($value) && isset($this->classMap[get_class($value)])) {
+            $this->classMap[get_class($value)]($this, $value);
+        } elseif (is_callable($value)) {
+            $value($this);
         } elseif (is_null($value)) {
             // noop
         } elseif (is_array($value)) {
@@ -142,7 +148,11 @@ class Writer extends XMLWriter {
 
         } elseif (is_object($value)) {
 
-            throw new InvalidArgumentException('The writer cannot serialize objects of type: ' . get_class($value));
+            throw new InvalidArgumentException('The writer cannot serialize objects of class: ' . get_class($value));
+
+        } else {
+
+            throw new InvalidArgumentException('The writer cannot serialize values of type: ' . gettype($value));
 
         }
 
--- a/tests/Sabre/Xml/WriterTest.php
+++ b/tests/Sabre/Xml/WriterTest.php
@@ -323,6 +323,49 @@ HI;
 
         $this->assertEquals($output, $this->writer->outputMemory());
 
+    }
+
+    function testCallback() {
+
+        $this->compare([
+            '{http://sabredav.org/ns}root' => function(Writer $writer) {
+                $writer->text('deferred writer');
+            },
+        ], <<<HI
+<?xml version="1.0"?>
+<s:root xmlns:s="http://sabredav.org/ns">deferred writer</s:root>
+
+HI
+        );
+
+    }
+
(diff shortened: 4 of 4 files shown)


staabm:
Any chanche we can get the classmap beeing used for reading and writing?

evert:
That wouldn't make a lot of sense, because during reading we don't know yet which class needs to be created. All we have is the element name.

staabm:
Oh, obviously you are right ;-)

staabm:
one thing which bugs me...

```php
class ErpOrderService extends Sabre\Xml\Service {
    public $elementMap = array(
        '{http://b2bOrder.complex.de}order_response' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
            return $popoDeserializer($reader, new OrderResponse(), XMLNS_ORDER);
        },
        '{http://b2bOrder.complex.de}order_status' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderStatus(), XMLNS_ORDER);
        },
    );
}
```

is not valid php, as I cannot use closures on the public field here

staabm:
ok, wil change it to 

```php
$erpOrderService = new Sabre\Xml\Service();
$erpOrderService->elementMap = array(
    '{'. XMLNS_ORDER .'}order_response' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderResponse(), XMLNS_ORDER);
    },
    '{'. XMLNS_ORDER .'}order_status' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderStatus(), XMLNS_ORDER);
    },
);
```

like the Reader/Writer examples

staabm:
... progress...

```php
$xml = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <order_response xmlns="http://b2bOrder.complex.de">
 <request_id>4711</request_id>
 <order_id>54752</order_id>
    <order_status xmlns="http://b2bOrder.complex.de">
         <status>INITIALIZED</status>
    </order_status>
 <success>true</success>
 </order_response>';

$popoDeserializer = function(Sabre\Xml\Reader $reader, $popo, $namespace)
{
    if ($reader->isEmptyElement) {
        throw new Exception(__CLASS__ . ' does not support empty elements');
    }

    $reader->read();
    do {

        if ($reader->nodeType === Sabre\Xml\Reader::ELEMENT && $reader->namespaceURI == $namespace) {

            $popo->{$reader->localName} = $reader->parseCurrentElement()['value'];
        } else {
            $reader->read();
        }
    } while ($reader->nodeType !== Sabre\Xml\Reader::END_ELEMENT);

    $reader->read();

    return $popo;
};

$popoSerializer = function(Sabre\Xml\Writer $writer, $popo, $namespace)  {
    foreach(get_object_vars($popo) as $key => $val) {
        $writer->writeElement('{'. $namespace .'}' . $key, $val);
    }
};

class OrderResponse
{
    public $request_id;
    public $order_id;
    public $status_status;
    public $success;
}

class OrderStatus
{
    public $status;
}

define("XMLNS_ORDER", "http://b2bOrder.complex.de");

$erpOrderService = new Sabre\Xml\Service();
$erpOrderService->namespaceMap[XMLNS_ORDER] = '';
// how to read xml into objects
$erpOrderService->elementMap = array(
    '{'. XMLNS_ORDER .'}order_response' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderResponse(), XMLNS_ORDER);
    },
    '{'. XMLNS_ORDER .'}order_status' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderStatus(), XMLNS_ORDER);
    },
);
// how to create xml from obejects
$erpOrderService->classMap = array(
    'OrderResponse' => function(Sabre\Xml\Writer $writer, $orderResponse) use ($popoSerializer) {
        /** @var $orderResponse OrderResponse */
        $popoSerializer($writer, $orderResponse, XMLNS_ORDER);
    },
    'OrderStatus' => function(Sabre\Xml\Writer $writer, $orderStatus) use ($popoSerializer) {
        /** @var $orderStatus OrderStatus */
        $popoSerializer($writer, $orderStatus, XMLNS_ORDER);
    }
);

$reader = $erpOrderService->getReader();
$reader->xml($xml);
$orderResp = $reader->parse()['value'];
var_dump($orderResp);
$writer = $erpOrderService->getWriter();
$writer->openMemory();
$writer->setIndent(true); // pretty print for debugging
$writer->write($orderResp);
var_dump($writer->outputMemory());
```

output

```sh
object(OrderResponse)#11 (5) {
  ["request_id"]=>
  string(4) "4711"
  ["order_id"]=>
  string(5) "54752"
  ["status_status"]=>
  NULL
  ["success"]=>
  string(4) "true"
  ["order_status"]=>
  object(OrderStatus)#12 (1) {
    ["status"]=>
    string(11) "INITIALIZED"
  }
}
string(205) "<:request_id xmlns="http://b2bOrder.complex.de">4711</:request_id>
<:order_id>54752</:order_id>
<:status_status/>
<:success>true</:success>
<:order_status>
 <:status>INITIALIZED</:status>
</:order_status>
"
```

except the `:` in the written xml, it looks fine ;)

evert:
> as I cannot use closures on the public field here

Alternatively you could also have set them in the constructor.

> except the : in the written xml, it looks fine ;)

What's with that `:`? That doesn't look right!

> in case of having a external deserializer it would make sense to have the serializer emit the whole element... wdyt? (in case we agree here, I cannot use the service->write as the root element would then be emitted twice

Personally I would not break the rule, but if you think you have good reasons to do so, you should ;)

staabm:
> What's with that :? That doesn't look right!

I dont know why. The above code is runnable, maybe it helps to repro the problem

staabm:
I guess the `:` are caused by the line `$erpOrderService->namespaceMap[XMLNS_ORDER] = '';`.

I do this to not get a namespace (inspired by the docs from http://sabre.io/xml/writing/)

evert:
Can you try `null` instead of `''` ?

staabm:
@evert you are right, we need `null` instead of `''`.

send a PR which fixes the example code.
https://github.com/fruux/sabre.io/pull/61

It also fixes another error with the code.

So after all this we have a working code:

```php
$xml = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <order_response xmlns="http://b2bOrder.complex.de">
 <request_id>4711</request_id>
 <order_id>54752</order_id>
    <order_status xmlns="http://b2bOrder.complex.de">
         <status>INITIALIZED</status>
    </order_status>
 <success>true</success>
 </order_response>';

$popoDeserializer = function(Sabre\Xml\Reader $reader, $popo, $namespace)
{
    if ($reader->isEmptyElement) {
        throw new Exception(__CLASS__ . ' does not support empty elements');
    }

    $reader->read();
    do {

        if ($reader->nodeType === Sabre\Xml\Reader::ELEMENT && $reader->namespaceURI == $namespace) {

            $popo->{$reader->localName} = $reader->parseCurrentElement()['value'];
        } else {
            $reader->read();
        }
    } while ($reader->nodeType !== Sabre\Xml\Reader::END_ELEMENT);

    $reader->read();

    return $popo;
};

$popoSerializer = function(Sabre\Xml\Writer $writer, $popo, $namespace)  {
    foreach(get_object_vars($popo) as $key => $val) {
        $writer->writeElement('{'. $namespace .'}' . $key, $val);
    }
};

class OrderResponse
{
    public $request_id;
    public $order_id;
    public $order_status;
    public $success;
}

class OrderStatus
{
    public $status;
}

define("XMLNS_ORDER", "http://b2bOrder.complex.de");

$erpOrderService = new Sabre\Xml\Service();
$erpOrderService->namespaceMap[XMLNS_ORDER] = null;
// how to read xml into objects
$erpOrderService->elementMap = array(
    '{'. XMLNS_ORDER .'}order_response' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderResponse(), XMLNS_ORDER);
    },
    '{'. XMLNS_ORDER .'}order_status' => function (Sabre\Xml\Reader $reader) use ($popoDeserializer) {
        return $popoDeserializer($reader, new OrderStatus(), XMLNS_ORDER);
    },
);
// how to create xml from obejects
$erpOrderService->classMap = array(
    'OrderResponse' => function(Sabre\Xml\Writer $writer, $orderResponse) use ($popoSerializer) {
        /** @var $orderResponse OrderResponse */
        $popoSerializer($writer, $orderResponse, XMLNS_ORDER);
    },
    'OrderStatus' => function(Sabre\Xml\Writer $writer, $orderStatus) use ($popoSerializer) {
        /** @var $orderStatus OrderStatus */
        $popoSerializer($writer, $orderStatus, XMLNS_ORDER);
    }
);

echo "<xmp>";
var_dump($xml);
$orderResp = $erpOrderService->parse($xml);
var_dump($orderResp);
var_dump($erpOrderService->write(get_class($orderResp), $orderResp));
```

do you think this kind of popo serializer/deserializer is something which should be shipped with sabre/xml or is it to specific for my current app?

evert:
I would definitely think it's a good idea to have this as a core feature btw!

evert:
Perhaps we can integrate this in the `Service` class with a function that takes a class name and xml element name and automatically registers both the serializer and deserializer.

staabm:
Agree that having it in the Service class is a good idea. Its the place where this kind of api-sugar should happen. PR incoming

PR #62, opened 2015-12-07
added a new namespaceAware() deserializer function

staabm:
didnt found the comments in which we discussed this topic therefore starting from scratch.
hopefully all points are adressed.

### DIFF
--- a/lib/functions.php
+++ b/lib/functions.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Sabre\Xml\Deserialize;
+
+use Sabre\Xml\Reader;
+
+/*
+ * Deserializes Xml Elements of the given Reader into a array.
+ * The namespace of elements which match the given namespace are stripped.
+ *
+ * @param Xml\Reader $reader
+ * @param string $namespace
+ *
+ * @return array
+ */
+function namespaceAware(Reader $reader, $namespace)
+{
+    // If there's no children, we don't do anything.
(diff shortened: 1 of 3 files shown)


lib/functions.php
```diff
@@ -0,0 +1,44 @@
...
+/*
+ * Deserializes Xml Elements of the given Reader into a array.
+ * The namespace of elements which match the given namespace are stripped.
+ *
+ * @param Xml\Reader $reader
+ * @param string $namespace
+ *
+ * @return array
+ */
+function namespaceAware(Reader $reader, $namespace)
```
staabm: feel free to propose a better name.. couldnt get better then this right now

evert:
I think I want to change the behavior of the function a little bit. Instead of :

`Sabre\Xml\Deserialize\namespaceAware`, I think I just want to call it `Sabre\Xml\Deserializer\keyValue`.

By default I want to give it the same behavior as `Sabre\Xml\Element\KeyValue`, but if you specify the `$namespace` argument (which will be optional) it will strip the namespace, but _only_ from the elements within that namespace, and leave the others intact.

So basically, hold on a second. I will take your PR and use it as a basis for this =)

staabm:
Sounds great!

evert:
Oh I didn't realize that in your PR you're already retaining other elements outside of the specified namespace. Really awesome. In that case I'm only changing the function name and docblocks ;)

PR #66, opened 2015-12-09
added popo serializer/deserializer and Service shorthand

staabm:
added popo mapping on the Service class.

see enclosed test how it is expected to work.
its more or less the same as initially discussed in https://github.com/fruux/sabre-xml/pull/61

After we finished the impl. details I will iterate over the phpdocs a bit.

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -189,6 +189,30 @@ class Service {
 
     }
 
+    /**
+     * Maps the given value-object to the given rootElementName within the given namespace.
+     *
+     * The $rootElement must be specified in clark notation.
+     *
+     * The $className refernces a regular php class which must provide a __construct() without arguments.
+     * This methods creates instances of this class which later on picks up all xml child elements as public properties.
+     *
+     * @param string $rootElementName
+     * @param object $className
+     * @param string $namespace
+     */
+    function mapValueObject($rootElementName, $className, $namespace) {
+        if (!class_exists($className)) {
+            throw new \InvalidArgumentException('class "' . $className . '" does not exist');
+        }
+        $this->namespaceMap[$namespace] = null;
+        $this->elementMap['{' . $namespace . '}' . $rootElementName] = function(Reader $reader) use ($className, $namespace) {
+            return \Sabre\Xml\Deserializer\valueObject($reader, new $className(), $namespace);
+        };
+        $this->classMap[$className] = function(Writer $writer, $valueObject) use ($namespace) {
+            return \Sabre\Xml\Serializer\valueObject($writer, $valueObject, $namespace);
+        };
+    }
 
     /**
      * Parses a clark-notation string, and returns the namespace and element
@@ -212,6 +236,4 @@ class Service {
         ];
 
     }
-
-
 }
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -205,7 +205,7 @@ class Service {
         if (!class_exists($className)) {
             throw new \InvalidArgumentException('class "' . $className . '" does not exist');
         }
-        $this->namespaceMap[$namespace] = null;
+ 
         $this->elementMap['{' . $namespace . '}' . $rootElementName] = function(Reader $reader) use ($className, $namespace) {
             return \Sabre\Xml\Deserializer\valueObject($reader, new $className(), $namespace);
         };
--- a/lib/deserializers.php
+++ b/lib/deserializers.php
@@ -161,3 +161,31 @@ function elementList(Reader $reader, $namespace = null) {
     return $values;
 
 }
+
+/**
+ * @param Reader $reader
+ * @param object $valueObject
+ * @param string $namespace
+ * @return null|$valueObject
+ */
+function valueObject(Reader $reader, $valueObject, $namespace) {
+    if ($reader->isEmptyElement) {
+        $reader->next();
+        return null;
+    }
+
+    $reader->read();
+    do {
+
+        if ($reader->nodeType === Reader::ELEMENT && $reader->namespaceURI == $namespace) {
+
+            $valueObject->{$reader->localName} = $reader->parseCurrentElement()['value'];
+        } else {
+            $reader->read();
+        }
+    } while ($reader->nodeType !== Reader::END_ELEMENT);
+
+    $reader->read();
+
+    return $valueObject;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ ChangeLog
 1.3.0 (????-??-??)
 ------------------
 
+* The `Service` class adds a new `mapValueObject` method which provides basic
+  capabilities to map between ValueObjects and XML.
 * #61: You can now specify serializers for specific classes, allowing you
   separate the object you want to serialize from the serializer. This uses the
   `$classMap` property which is defined on both the `Service` and `Writer`.
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,10 @@
         "psr-4" : {
             "Sabre\\Xml\\" : "lib/"
         },
-        "files": ["lib/deserializers.php"]
+        "files": [
+            "lib/deserializers.php",
+            "lib/serializers.php"
+        ]
     },
     "bin" : [
     ],
--- a/lib/serializers.php
+++ b/lib/serializers.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Sabre\Xml\Serializer;
+
+use Sabre\Xml\Writer;
+
+/**
+ * This file provides a number of 'serializer' helper functions.
+ * These can be used to easily specify custom serializers for specific
+ * PHP objects/values.
+ */
+
+/**
+ * @param Writer $writer
+ * @param object $valueObject
+ * @param string $namespace
+ */
+function valueObject(Writer $writer, $valueObject, $namespace) {
+    foreach (get_object_vars($valueObject) as $key => $val) {
+        $writer->writeElement('{' . $namespace . '}' . $key, $val);
+    }
+}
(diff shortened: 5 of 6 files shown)


lib/Service.php
```diff
@@ -189,6 +189,27 @@ function write($rootElementName, $value, $contextUri = null) {
 
     }
 
+    /**
+     * Maps the given plain-old-php-object (popo) to the given rootElementName within the given namespace.
```
evert: I'm kind of wonderering if we should go with `VO` or even `ValueObject` instead of `Popo`. I got what it meant because of Pojo, but I'm not certain how well known that terminology is in PHP... What do you think?

lib/Service.php
```diff
@@ -189,6 +189,27 @@ function write($rootElementName, $value, $contextUri = null) {
...
+     *
+     * The $rootElement must be specified in clark notation.
+     *
+     * @param string $rootElementName
+     * @param object $popo
+     * @param string $namespace
+     */
+    function mapPopo($rootElementName, $popo, $namespace) {
+        if (!is_object($popo)) {
```
evert: The first thing that popped in my mind is ... should it be based on classnames instead of objects? It seems a bit odd to me to pass an object instance that's then cloned...  Thoughts?

lib/deserializers.php
```diff
@@ -161,3 +161,31 @@ function elementList(Reader $reader, $namespace = null) {
...
+function popoObject(Reader $reader, $popo, $namespace) {
+    if ($reader->isEmptyElement) {
+        $reader->next();
+        return null;
+    }
+
+    $reader->read();
+    do {
+
+        if ($reader->nodeType === Reader::ELEMENT && $reader->namespaceURI == $namespace) {
```
evert: Future idea: add the ability to parse annotations and let people create classes like this:

```php

class MyVO {

   /**
    * @map {http://namespace}elem
    */
   public $foo;

}
```

lib/Service.php
staabm: I am used to the term popo from symfony or propel, but I am also ok with using a more 'pattern-oriented' term like ValueObject. 

I guess both terms are only known to experienced php programmers, so we can decide whatever we like best.

lib/Service.php
staabm: Passing in the class would imply a `__construct()` without arguments therefore I thougt using a "template-object" would be a good idea since you could also prefill object properties as required.

lib/deserializers.php
staabm: Yeah, I also thought about such things... I dont like annotations that much because they add a magic layer which needs to be documented very properly. Also it mixes the serialize/deserialize logic into the ValueObjects which is the opposite direction this PR tries to aim for.

Maybe this will improve in php-src future in case annotations get a 1st class citizen in php.

lib/Service.php
evert: I think I would lean towards ValueObject, it also sounds a bit better and I feel it requires a bit less explanation...

lib/deserializers.php
staabm: Having annotations also opens the door for mapping xml attributes somewhere. Atm we dont support the at all in this popo cases

lib/Service.php
evert: Maybe we can do both? I feel it's more natural to allow just classes and force people to create their VO's without constructors, but I can definitely see an advantage to both

lib/deserializers.php
evert: Could this function use use `keyValue` instead? It's one more layer of indirection, but it makes the implementation of popoObject a lot simpler (which can benefit people reading the source).

lib/Service.php
staabm: Ok, so how should the method be named?

`mapVo` or `mapValueObject`? I prefer the latter

lib/Service.php
staabm: I am fine with it. Do you have a use case for the classnames already or is it just hypothetical?

In case there is not use case yet I would prefer waiting for one

lib/Service.php
evert: I agree.. was just gonna say that another benefit of using the term VO means that we can fully spell it out (as opposed to PlainOldPhpObject ;))

lib/deserializers.php
staabm: Its one indirection and another short living temp. Structure.. In case you prefer it I am fine with it. We can restore to what we have right now in case it get a perf bottleneck

lib/Service.php
evert: It's a bit more of a design decision, because whatever we allow is what people will have to do... and elsewhere in the source generally rely on PHP class names (such as `$classMap` and `$elementMap`) so it would be internally consistent.

So between the two that would be my first choice. Not only for consistency, but it also plays a bit nicer with the idea that 'objects should be fully initialized after creation', and you can argue that a 'template object' is not a fully constructed object.

staabm: Ok, will go for classes then. Atm this concept also works in my use case

lib/Service.php
staabm: Will update tomorrow

lib/deserializers.php
evert: Lets do it then =)

staabm: Absolutely.

evert: Actually... I changed my mind!

I thought using `keyValue` this function would become super short, but that's not true... you would still need to loop through the result of `keyValue` and filter out the namespaced elements... which sucks.

staabm: Okidoki

evert: lol

staabm:
here we go

lib/Service.php
```diff
@@ -189,6 +189,29 @@ function write($rootElementName, $value, $contextUri = null) {
...
+     *
+     * @param string $rootElementName
+     * @param object $className
+     * @param string $namespace
+     */
+    function mapValueObject($rootElementName, $className, $namespace) {
+        if (!class_exists($className)) {
+            throw new \InvalidArgumentException('class "' . $className . '" does not exist');
+        }
+        $this->elementMap['{' . $namespace . '}' . $rootElementName] = function(Reader $reader) use ($className, $namespace) {
```
staabm: I dropped the `$this->namespaceMap[$namespace]=null` expression here, because it had side effects.
I lead to all elements of this namespaces beeing represented by its localName instead of clark-notated (also for this not "registered" via `mapValueObject`).

After dropping the line we are still green, so nothing important ;)

staabm: strike that. after I recognized that I looked at the wrong repos on travis I realised the line is still necessary.

therefore the question... what do you think about this sideeffect? any idea how to get arround it?

lib/deserializers.php
staabm: time is up. pushed without the additional indirection ;)

evert:
On the first read this looks pretty great! No comments

lib/Service.php
evert: I'm a bit confused because this thread is on an 'outdated diff' discussing a line that wasn't in that diff...  can you point me in the right direction?

staabm: This is the Line in the current diff https://github.com/fruux/sabre-xml/pull/66/files#diff-de76feda74829916defbb14d089c7c08R208

lib/Service.php
```diff
@@ -189,6 +189,30 @@ function write($rootElementName, $value, $contextUri = null) {
...
+     *
+     * @param string $rootElementName
+     * @param object $className
+     * @param string $namespace
+     */
+    function mapValueObject($rootElementName, $className, $namespace) {
+        if (!class_exists($className)) {
+            throw new \InvalidArgumentException('class "' . $className . '" does not exist');
+        }
+        $this->namespaceMap[$namespace] = null;
```
staabm: This line has a side effect on all elements of the given namespace not only the one named root-element

evert: oh yea that does seem like a big problem! Why is it in there in the first place?

staabm: Otherwise we would get some aliases in the xml

evert: Well I think I would just want to encourage people to always (manually) set their preferred namespace map.

I would also suggest to always use some prefix, and not default to no prefix at all. The fallback with aliases is specifically added because it generates valid (but somewhat ugly) xml, but you, as the xml author, need to pick your preferred namespace prefix (even if that prefix is empty).

So yea this is definitely by design. We

evert: We _could_ add a new feature that sets the default namespace to `null`, but _only_ for the root element. So this needs to be detected, and happen outside of of the serializer functions

staabm: How do I need to define the namespace, so the xml us mapped to something which I defined myself?

evert: as a user of `Sabre\Xml\Service`, you would set:

    $server->namespaceMap['uri'] = 'prefix';

This is how sabre/dav does it:

https://github.com/fruux/sabre-dav/blob/master/lib/DAV/Xml/Service.php

evert:
So yea, I think the last two tasks are to remove that `$this->namespaceMap[$namespace] = null;` line and add the missing docblocks =)

staabm:
After removing the line it wont generate the xml I want it to render

evert:
Just add the line to the unittest then ;) You can set the namespace prefixes from outside the Service object, and that's also exactly how you are supposed to do it.

staabm:
Uff thats too easy... :-))

staabm:
@evert not sure you got the last ping.. This is ready to land

evert:
The thing I was still waiting for is better php docblocks. But I can merge this now and I'd be happy to address that before a release. Really great work :+1:

staabm:
Ohh sorry missed the docs :-)

PR #68, opened 2015-12-12
Refactoring the default serialization behavior.

evert:
This is a work in progress

### DIFF
--- a/lib/serializers.php
+++ b/lib/serializers.php
@@ -2,12 +2,15 @@
 
 namespace Sabre\Xml\Serializer;
 
+use InvalidArgumentException;
 use Sabre\Xml\Writer;
+use Sabre\Xml\XmlSerializable;
 
 /**
  * This file provides a number of 'serializer' helper functions.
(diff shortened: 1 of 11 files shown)


lib/serializers.php
```diff
@@ -25,3 +60,165 @@ function valueObject(Writer $writer, $valueObject, $namespace) {
...
+        // A callback
+        $value($writer);
+
+    } elseif (is_null($value)) {
+
+        // nothing!
+
+    } elseif (is_array($value)) {
+
+        reset($value);
```
staabm: wont `foreach` automatically reset the array?

staabm: see http://php.net/manual/en/control-structures.foreach.php
```
Note:

In PHP 5, when foreach first starts executing, the internal array pointer is automatically reset to the first element of the array. This means that you do not need to call reset() before a foreach loop. 
```

evert: Huh weird.. I don't know why `reset` is in there. Perhaps some leftover from an earlier iteration of this. Removing it.

PR #71, opened 2015-12-21
A number of improvements to the new VO mapping system.

evert:
* New docblocks
* Updated some method signatures

cc: @staabm


staabm:
didnt test it yet, but looks good for now. thx, it simplifies things a bit.

staabm:
@evert thinking a bit further to improve the feature:

wouldnt it work to have a `$service->writeObject($object)` method, which can write any object in case it is of a class previously registered via `mapValueObject`?

that way I wouldnt need to repeat the clark-notation at write-time and need only the declaration...?

staabm:
Hmm I guess what I am after is `$service->getWriter()->write($object)`  instead of `$service->write($object)`.

If this is the intended way, the api needs some polish IMO. ValueObjects are registered at `Service` level but can only be written as-is on `Writer` level. maybe just need a shorthand `$service->writeObject()` as mentioned above.

staabm:
ATM I am doing something like

```php
class Myservice extends Service {
    const XMLNS_ORDER = "http://myns/order";

    private $clarkMap = array();

    public function __construct() {

        // make writer produce non-namespaced elements
        $this->namespaceMap[self::XMLNS_ORDER] = null;

        $this->mapValueObject('{'. self::XMLNS_ORDER .'}cancel_order_request', 'CancelOrderRequest');

        parent::__construct();
    }

    public function mapValueObject($elementName, $className) {
        parent::mapValueObject($elementName, $className);
        $this->clarkMap[$className] = $elementName;
    }

    public function writeObject($object) {
        if (!isset($this->clarkMap[get_class($object)])) throw new Exception('"'. get_class($object) .'" is not a registered value object class');

        return $this->write($this->clarkMap[get_class($object)], $object);
    }
}
```

to get arround this

evert:
I understand the issue, but:

    $service->getWriter()->write($object)

is not supposed to work. It it does, we have a bug. Consider this scenario:

```php

<?php
class VO1 {
   public $vo2;
}

class VO2 {
   public $foo = 'bar';
}

$vo1 = new VO1();
$vo1->vo2 = new VO2();

$server = new Service();
$server->mapValueObject('{}vo1', 'VO1');
$server->mapValueObject('{}vo2', 'VO2');
$service->write($vo1);
```

Should output:

```xml
<?xml version="1.0"?>
<vo1>
   <vo2>
      <foo>bar</foo>
   </vo2>
</vo1>
```

staabm:
After testing a bit more I came to the same conclusion. It doesnt work with `$service->getWriter()->write($object)` either :-).

What do you think about a `writeObject` like method?
I am fine with having this kind of feature at my app level but it feels like a good addition for the valueObject based approach

evert:
I think it's a good idea. I'm creating a PR now but I'll call it `writeVO` instead =)

PR #74, opened 2015-12-23
Fix infinite loop on unknown elements in VO's.

### LINKED ISSUE #73: Endless loop in Service->parse
using the following ...

... ValueObject
```php
class ErpOrderResponse {
    public $request_id;
    public $order_id;
    // because this property is missing, things run endless.
    // uncommenting the line will make things work again
    // public $status;
    public $version;
    public $success;
}
```

... XML
```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<order_response xmlns="http://clxERP/order">
    <request_id>ErpService:Wed, 23 Dec 2015 13:06:23 +0100</request_id>
    <order_id>56652</order_id>
    <status>CANCELLED</status>
    <version>9</version>
    <success>true</success>
</order_response>
```

... Service
```php
class ErpMarshall extends Sabre\Xml\Service {
    const XMLNS_ORDER = "http://clxERP/order";

    private $clarkMap = array();

    public function __construct() {

        // make writer produce non-namespaced elements
        $this->namespaceMap[self::XMLNS_ORDER] = null;

        $this->mapValueObject('{'. self::XMLNS_ORDER .'}cancel_order_request', 'ErpCancelOrderRequest');
        $this->mapValueObject('{'. self::XMLNS_ORDER .'}order_response', 'ErpOrderResponse');
    }

    public function mapValueObject($elementName, $className) {
        parent::mapValueObject($elementName, $className);
        $this->clarkMap[$className] = $elementName;
    }

    public function writeObject($object) {
        if (!isset($this->clarkMap[get_class($object)])) throw new Exception('"'. get_class($object) .'" is not a registered value object class');

        return $this->write($this->clarkMap[get_class($object)], $object);
    }
}
```

I guess this service class can be simplified further to get a minimal testcase.

things run endless when you do a parse like
```php
$marshall = new ErpMarshall();
var_dump($marshall->parse($xml));
```

uncommenting the property `status` in the `ErpOrderResponse` class makes things work as expected.
I agree its a programming mistake, but I feel we need to improve the DX.

As I run the code on the CLI it also never timeouts, which is a different problem ;).

evert:
Fixes #73

### DIFF
--- a/lib/deserializers.php
+++ b/lib/deserializers.php
@@ -189,6 +189,9 @@ function valueObject(Reader $reader, $className, $namespace) {
 
             if (property_exists($valueObject, $reader->localName)) {
                 $valueObject->{$reader->localName} = $reader->parseCurrentElement()['value'];
+            } else {
+                // Ignore property
+                $reader->next();
             }
         } else {
             $reader->read();
--- a/tests/Sabre/Xml/DeserializersTest.php
+++ b/tests/Sabre/Xml/DeserializersTest.php
@@ -74,7 +74,7 @@ XML;
         $reader->xml($input);
         $reader->elementMap = [
             '{urn:foo}foo' => function(Reader $reader) {
-                return valueObject($reader, TestVo::class, 'urn:foo');
+                return valueObject($reader, 'Sabre\\Xml\\Deserializer\\TestVo', 'urn:foo');
             }
         ];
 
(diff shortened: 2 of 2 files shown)


staabm:
Fix will be verified on monday, when I am back on my workstation. Thx so far for your quick response

PR #76, opened 2015-12-23
Added writeValueObject.

evert:
cc: @staabm
needs tests

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -233,6 +233,32 @@ class Service {
     }
 
     /**
+     * Writes a value object.
+     *
+     * This function largely behaves similar to write(), except that it's
+     * intended specifically to serialize a Value Object into an XML document.
+     *
+     * The ValueObject must have been previously registered using either
+     * mapValueObject() or the $classMap property.
+     *
+     * @param object $object
+     * @param string $contextUri
+     * @return void
+     */
+    function writeValueObject(object $object, $contextUri = null) {
+
+        if (!isset($this->classMap[get_class($object)]))
+            throw new \InvalidArgumentException('"'. get_class($object) .'" is not a registered value object class. Register your class with either mapValueObject or $classMap.');
+        }
+        return $this->write(
+            $this->classMap[get_class($object)],
+            $object,
+            $contextUri
+        );
+
+    }
+
+    /**
      * Parses a clark-notation string, and returns the namespace and element
      * name components.
      *
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -230,6 +230,7 @@ class Service {
         $this->classMap[$className] = function(Writer $writer, $valueObject) use ($namespace) {
             return \Sabre\Xml\Serializer\valueObject($writer, $valueObject, $namespace);
         };
+        $this->valueObjectMap[$className] = $elementName;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


lib/Service.php
```diff
@@ -233,6 +233,32 @@ function mapValueObject($elementName, $className) {
...
+     * @param string $contextUri
+     * @return void
+     */
+    function writeValueObject(object $object, $contextUri = null) {
+
+        if (!isset($this->classMap[get_class($object)]))
+            throw new \InvalidArgumentException('"'. get_class($object) .'" is not a registered value object class. Register your class with either mapValueObject or $classMap.');
+        }
+        return $this->write(
+            $this->classMap[get_class($object)],
```
staabm: Write req. 1st arg to be a clark notation, not a callabe

lib/Service.php
```diff
@@ -233,6 +233,32 @@ function mapValueObject($elementName, $className) {
...
+     * intended specifically to serialize a Value Object into an XML document.
+     *
+     * The ValueObject must have been previously registered using either
+     * mapValueObject() or the $classMap property.
+     *
+     * @param object $object
+     * @param string $contextUri
+     * @return void
+     */
+    function writeValueObject(object $object, $contextUri = null) {
```
staabm: Is `object` a valid typehint?

staabm:
Thx for picking it up. I like the pace we iterate on this lib. Thx for doing such a great job :100:

lib/Service.php
evert: it wasn't =( sadly

evert:
Thanks for all the web points!

tests/Sabre/Xml/ServiceTest.php
```diff
@@ -231,10 +231,20 @@ function testMapValueObject() {
 
         $this->assertEquals($expected, $order);
 
-        $writtenXml = $orderService->write('{http://sabredav.org/ns}order', $order);
+        $writtenXml = $orderService->writeValueObject($order);
```
staabm: Shouldn't we test both api calls and dont assume impl details (that those are equivalent). We should assert those are equivalent instead

lib/Service.php
```diff
@@ -230,6 +230,33 @@ function mapValueObject($elementName, $className) {
         $this->classMap[$className] = function(Writer $writer, $valueObject) use ($namespace) {
             return \Sabre\Xml\Serializer\valueObject($writer, $valueObject, $namespace);
         };
+        $this->valueObjectMap[$className] = $elementName;
```
staabm: Note related to this PR: do we need to make sure (or help preventing users people shooting in there feet)?

- check whether a classMapping already exists before mapping the value object
- check whether the elementMapping already exists before kaüping the value object

evert: Idk! I think overwriting them is a legitimate use-case as well :/

staabm: Ok, then we should add tests for the case

evert: Or maybe not. Overwriting *inside* of Writer or Reader makes a ton of sense, but maybe not in Service. Could be good to throw an exception there

staabm: Thats what I meant ..

evert: Sorry I'm flip-flopping again. I think we need to just let people overwrite these if they want to

evert: The use-case is if an application provides a service, and a module alters the standard behavior the application provides. I don't think that that's entirely far-fetched.

staabm: :-)

PR #77, opened 2015-12-24
Automatically support VO properties that are specified as arrays.

evert:
This is a feature for VO's such as this:

    class VO {

       public $foo = [];

    }

Specifying the empty array as the default will cause the parser to
assume that foo may appear more than once and append elements to the
array.

cc: @staabm

### DIFF
--- a/lib/deserializers.php
+++ b/lib/deserializers.php
@@ -182,13 +182,19 @@ function valueObject(Reader $reader, $className, $namespace) {
         return $valueObject;
     }
 
+    $defaultProperties = get_class_vars($className);
+
     $reader->read();
     do {
 
         if ($reader->nodeType === Reader::ELEMENT && $reader->namespaceURI == $namespace) {
 
             if (property_exists($valueObject, $reader->localName)) {
-                $valueObject->{$reader->localName} = $reader->parseCurrentElement()['value'];
+                if (is_array($defaultProperties[$reader->localName])) {
+                    $valueObject->{$reader->localName}[] = $reader->parseCurrentElement()['value'];
+                } else {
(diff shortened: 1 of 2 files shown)


lib/deserializers.php
```diff
@@ -182,13 +182,19 @@ function valueObject(Reader $reader, $className, $namespace) {
...
+    $defaultProperties = get_class_vars($className);
+
     $reader->read();
     do {
 
         if ($reader->nodeType === Reader::ELEMENT && $reader->namespaceURI == $namespace) {
 
             if (property_exists($valueObject, $reader->localName)) {
-                $valueObject->{$reader->localName} = $reader->parseCurrentElement()['value'];
+                if (is_array($defaultProperties[$reader->localName])) {
```
staabm: We could also check if `valueObject->{reader->localName}` is an array and append to instead. Then we wouldnt need to retrieve the class-vars beforehand, wdyt?

staabm:
Nice idea!

lib/deserializers.php
evert: There's a possibility that parseCurrentElement returns an array, for certain property mappings. So the easiest was to use to grab this list beforehand.

staabm:
@evert what do you think about doing the same for `object` type?

```php
class VO1 {

   public $foo = new SubVo();
}

class SubVo {
  public $name;
  public $xyz;
}
```

in this case we would not create a new object in the deserializer but pickup the one already present in the object-property...?

evert:
Well, I don't think that's valid PHP ;)

More importantly, you can actually achieve this with the mapValueObject function already.

PR #82, opened 2015-12-30
Cache namespaces parsed from clark notation

icewind1991:
When serializing large xml documents the cost parsing of the namespaces from clark notation becomes a significant portion (>10%) of the serialization cost.
Since most xml documents will only use a fairly small number of xml elements we're better of keeping the parsed version of the namespace/element around for reuse.

This saves doing a regex (and some memory allocations) for each element we serialize

[comparison](https://blackfire.io/profiles/compare/4449f8f7-abbc-4b7d-89a4-a3770e75b053/graph)

cc @evert

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -266,16 +266,21 @@ class Service {
      * @return array
      */
     static function parseClarkNotation($str) {
+        static $cache = [];
 
-        if (!preg_match('/^{([^}]*)}(.*)$/', $str, $matches)) {
-            throw new \InvalidArgumentException('\'' . $str . '\' is not a valid clark-notation formatted string');
+        if (!isset($cache[$str])) {
+
+            if (!preg_match('/^{([^}]*)}(.*)$/', $str, $matches)) {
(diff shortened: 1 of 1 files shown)


staabm:
This patch saves 30ms walltime, what is the total parsing walltime? 300ms?

Would it be possible to do the Caching in the `Service` class, so all calls to it can benefit from the cache?

evert:
I'm a bit new to blackfire. Is there a way to find out which script you used to generate the profile? It if it's all good and the benefit checks out, I do agree with @staabm that I think it's a bit better to move the cache to the function in the Service class, not in the least because it also means that the cache survives multiple runs.

You can simply declare a static property inside a function body like this:

```php
function foo() {

   static $cache;

}
```

evert:
Never mind on sharing the original script, I seen now that it was a whole webdav request. Looking into this a bit further

icewind1991:
Switched to using a static variable in the method.

Updated comparison: https://blackfire.io/profiles/compare/74200494-931e-436c-b367-1ec3fca31419/graph

evert:
Failing build is unrelated. Looks good!

PR #84, opened 2016-02-04
Fix function names in Deserializer/EnumTest

jakobsack:
The old function names might irritate people.


evert:
People such as yourself? :P

PR #89, opened 2016-03-04
added failling test for elements with "circular" dependencies

staabm:


staabm:
@evert any insights why this test is failling?

evert:
> It feels like the value-object feature should automatically pick-up attributes when equal named properties exist on the object (in case they are unique and no equal-named sub-element exist).

That sounds exactly why the test is failing. Automatically mapping attributes to properties is imho not the greatest solution, because we can't easily figure out what to do when we serialize the objects again. I think it's important that the VO stuff can operate bi-directional.

> alternative2: add a notion of "attribute-serializers/deserializers" additionally to our "element" based ones.

The atom package does this:

https://github.com/fruux/sabre-xml-atom/blob/master/lib/Service.php#L40

So yea I don't think it's a bad idea to add it to the standard list of serializers/deserializers. Clearly there's more than one person who needed this.

staabm:
In my use cse I need a mix of what the valueObject serializer provides in combination with the attribute reading atom does

evert:
In that case I would say you are stuck writing custom serializers/deserializers for those cases. I would say that that is by design :)

PR #90, opened 2016-03-12
Disable xdebug for composer install to speedup build

staabm:
We dont use xdebug features in travis, therefore disabling it which speeds up build considerably.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,6 @@ script:
   - ./bin/phpunit --configuration tests/phpunit.xml.dist
   - ./bin/sabre-cs-fixer fix . --dry-run --diff
 
-before_script: composer install
+before_script:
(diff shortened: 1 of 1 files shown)


staabm:
Will sqash the commits when back on a real wokrstation

staabm:
Hmm this offical way to disable the extension has problems with hhvm..

Need to adjust this ( and all the other PRs)

staabm:
ready to land

.travis.yml
```diff
@@ -19,4 +19,6 @@ script:
   - ./bin/phpunit --configuration tests/phpunit.xml.dist
   - ./bin/sabre-cs-fixer fix . --dry-run --diff
 
-before_script: composer install
+before_script:
+  - rm -f ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini
```
staabm: we cannot use the proper travis `phpenv` api, because it doesnt work with hhvm right now

PR #91, opened 2016-03-13
Drop hhvm on travis.

staabm:
Hhvm does not yet provide the classes required. Lets save some travis resources to protect the rainforest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,12 +4,9 @@ php:
   - 5.5
   - 5.6
   - 7
-  - hhvm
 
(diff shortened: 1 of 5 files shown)


evert:
I'd be more in favor of logging a bug against hhvm to get it to behave identically =) Just haven't gotten around to it yet :(

staabm:
Ok, let me take care of that

staabm:
lets see https://github.com/facebook/hhvm/issues/6913

PR #96, opened 2016-05-17
restore reader-context in early-return paths

staabm:
this are edge-cases but still those should restore the proper context of the current reader.

I didnt experience a concrete problems with this but just stumbled across it while reading the code.
Thats why no unit test provided.

Feel free to close this without merging it in case you think its not worth it.


evert:
this is great!

evert:
Actually... should have probably been solved with `finally` !

evert:
Oh well, we can do that later =)

PR #105, opened 2016-08-26
Updated keyValue deserializer

snebes:
XML that contains elements that have multiple occurrences are returned as an array of those elements, rather than just the last element in the list.

Given this XML:
```
<list>
    <meta>
        ...
    </meta>
    <report>
        <name/>
    </report>
    <report>
        <name/>
    </report>
    <report>
        <name/>
    </report>
    <report>
        <name/>
    </report>
</list>
```
The resulting object could be returned:
```
Sabre Object
(
    [meta] => Meta Object
        (
            ...
        )
    [report] => Array
        (
            [0] => Report Object
                (
                    [title] => Report Title
                )

            [1] => Report Object
                (
                    [title] => Report Title
                )

            [2] => Report Object
                (
                    [title] => Report Title
                )

            [3] => Report Object
                (
                    [title] => Report Title
                )

            [4] => Report Object
                (
                    [title] => Report Title
                )
```

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -81,15 +81,16 @@ function keyValue(Reader $reader, $namespace = null) {
                 $value = $reader->parseCurrentElement()['value'];
             }
 
-            if (is_array($values[$key])) {
-                $values[$key][] = $value;
-            }
-            elseif (array_key_exists($key, $values)) {
-                $tmp = $values[$key];
-                $values[$key] = [$tmp, $value];
+            if (!isset($values[$key])) {
+                $values[$key] = $value;
             }
             else {
-                $values[$key] = $value;
+                if (!is_array($values[$key])) {
+                    $tmp = $values[$key];
+                    $values[$key] = [$tmp];
+                }
+
+                $values[$key][] = $value;
             }
         } else {
             $reader->read();
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -51,7 +51,8 @@ use Sabre\Xml\Reader;
  * ];
  *
  * Attributes will be removed from the top-level elements. If elements with
- * the same name appear twice in the list, only the last one will be kept.
+ * the same name appear twice in the list, an array of those elements are
+ * returned.
  *
  *
  * @param Reader $reader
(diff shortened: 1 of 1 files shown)


evert:
Hi @snebes , the real answer here is: don't use KeyValue for this stuff.

The feature you're proposing might make sense for your use-case, but consider this:
`report` clearly is something that might appear 1 or more times. If it appears once, it will show up as it's bare value.. if it appears more than once, it will be an array of values.

The implication is that everywhere you are consuming this you need to now add if statements to handle both the array and non-array case. It also means that for cases where 1 element usually is expected, someone can construct an XML document that trips up the code that consumes the result of parsing.  In your example I can add a second `<meta>` and suddenly your code does weird things (or you need to assume *everywhere* KeyValue is used that it can *always* be a value, or array of values). Neither is good and neither is 'simple'.

There is a 'correct' solution to this however. The `keyValue` function is actually quite simple and for situations where the normal `keyValue` does not apply, it's extremely easy (and recommended!) to write your own.

Here's two solutions to the problem:

Solution #1 
-----------

```php

function listDeserializer(Reader $reader) {

   $result = [
      'meta' => null,
      'report' =>[],
   ];

   $namespace = '{http://blabla}';
   $elements = $reader->parseGetElements();
   foreach($elements as $element) {

        switch($element['name']) {

             case $namespace . "meta" :  
                $result['meta'] = $element['value'];
                break;
             case $namespace . "report" :
                $result['report'][] = $element['value'];
                break;
        } 

    }
    return $result;
  
}

$service->elementMap['{http://blabla}list'] = 'listDeserializer';

```

The above solution makes the result extremely predictable, even when there's 0 or 1 reports. I hope it's also still readable/simple enough.

Solution #2 
-----------

Value objects are similar to the keyValue deserializer, except they are class-based instead of array-based and allow you to hint that there's more than 1 element expected by setting a default value to an empty array.

```php

class List {

  public $meta;
  public $report = [];
   
}

$service->mapValueObject('{http://blabla}list','List');

```

snebes:
@evert Thank you for the feedback.  What you proposed is a great solution for a specific case, but the goal of the PR was to make a generic deserializer use-case for that situation.  I didn't expect it to get merged, but to start a conversation on a general solution for this use-case.

I agree that the need for condition checks on is_object/is_array/is_scalar are a blocker for this PR. I would fall victim to using these in my library as well.

evert:
Well, my take on this is that both solutions are 'generic enough'. Sure, with Solution 1 you need to write some code to correctly decode this; but the alternative is creating a function that configures a standard deserializer to treat 1 item as a 'one' value, and another as 'many'.

But I think the complexity of configuring such a deserializer rivals just writing a custom serialize for it. After all, writing code is a form of configuration.

I think if "configuring generic code to do x" gets close enough to "write specific code to do x", it's almost always better to do the second thing, unless you're a Java guy.

So if there's a viable generic solution to this problem, it should pass the test that it's actually simpler than a non-generic solution.

By the way, did you also consider solution 2? That _is_ actually a generic solution. You're mapping to classes, which give you a number of other benefits. Notably strict typing.

PR #110, opened 2016-10-07
Adds ability to stop reading XML using close() method

gijsstegehuis:


lib/Reader.php
```diff
@@ -22,6 +22,12 @@ class Reader extends XMLReader {
     use ContextStackTrait;
 
     /**
+     * Reading indicator
+     * @var bool
+     */
+    public $reading = false;
```
staabm: is this required to be public?

lib/Reader.php
```diff
@@ -322,4 +328,46 @@ function getDeserializerForElementName($name) {
...
+    /**
+     * Close the XMLReader input
+     * @return bool             Returns TRUE on success or FALSE on failure.
+     */
+    function close() {
+        if (parent::close()) {
+            $this->reading = false;
+        }
+
+        return $this->reading;
```
staabm: this should return whatever parent::close() returned, no?

staabm: I guess the same is true for all 3 methods.

lib/Reader.php
gijsstegehuis: Not required, but might be useful though.

staabm: please either describe a usefull case or reduce it to public.

lib/Reader.php
gijsstegehuis: Agree. It's fixed now.

lib/Reader.php
```diff
@@ -351,23 +347,15 @@ function open($uri, $encoding = null, $options = 0) {
...
      * Close the XMLReader input
      * @return bool             Returns TRUE on success or FALSE on failure.
      */
     function close() {
-        if (parent::close()) {
-            $this->reading = false;
-        }
-
-        return $this->reading;
+        return $this->reading = !parent::close();
```
staabm: I think this has to be 

```
$this->reading = false;
return  parent::close();
```

lib/Reader.php
gijsstegehuis: Changed it to protected in https://github.com/fruux/sabre-xml/pull/110/commits/548fc32d613e374429d13fe6172b7fa408a50893.

staabm: protected would be suitable when you can describe a usefull case in which you need it ;-).

if you can't -> private.

staabm: please describe a concrete example or use private.

gijsstegehuis: Changes it to private in https://github.com/fruux/sabre-xml/pull/110/commits/c8ca514ade1de7553c4e40797aaddf588a012652.

gijsstegehuis:
@staabm build is failing because of some Composer dependency issue thats not related to this PR.

lib/Reader.php
```diff
@@ -22,6 +22,12 @@ class Reader extends XMLReader {
     use ContextStackTrait;
 
     /**
+     * Reading indicator
+     * @var bool
+     */
+    private $reading = false;
```
evert: I think I prefer the negative here. `$stopped` instead of `$reading`.

lib/Reader.php
```diff
@@ -322,4 +328,35 @@ function getDeserializerForElementName($name) {
...
     }
 
+    /**
+     * Set the URI containing the XML to parse
+     * @param  string  $uri
+     * @param  string  $encoding
+     * @param  int $options
+     * @return bool            Returns TRUE on success or FALSE on failure
+     */
+    function open($uri, $encoding = null, $options = 0) {
```
evert: I think the true end-goal of your patch is not track when `open` is called, because that would be much simpler. I think what you want is being able to stop `parseInnerTree()`. Therefore I think you should also just set `$this->stopped = false` in that function, not in `open`/`close`.

lib/Reader.php
```diff
@@ -322,4 +328,35 @@ function getDeserializerForElementName($name) {
...
+     */
+    function xml($source, $encoding = null, $options = 0) {
+        return $this->reading = parent::xml($source, $encoding, $options);
+    }
+
+    /**
+     * Close the XMLReader input
+     * @return bool             Returns TRUE on success or FALSE on failure.
+     */
+    function close() {
```
evert: I'm curious if there's a need for override `close`. Could you just add a separate `->stop` function instead? The tie in with `close` seems kind of unimportant, but I could be wrong.

evert:
I'm closing this ticket because it's been a while since the last comment.

If this is indeed still an issue, feel free to comment here so we can continue discussing.

PR #114, opened 2017-01-02
Dependency sabre/uri for PHP5

marknl:
sabre/xml 2.0 can only be used on php7. So when you use php5, you need to fall back to sabre/xml 1.5.x, that's fine.

sabre/xml 2.0 requires: sabre/uri: >=1.0,<3.0.0
sabre/xml 1.5.x requires: sabre/uri: >=1.0,<3.0.0
sabre/uri >=1.0,<3.0.0 is currently version 2.1.0 which is also php7 only.

So installing sabre/xml 1.5.x will install the incorrect sabre/uri version. For php 5 sabre 1.5 needs to use sabre/uri 1.2.x.

In this fix I used  <1.3 assuming this will always have php 5 support.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,7 @@
         "ext-xmlreader" : "*",
         "ext-dom" : "*",
         "lib-libxml" : ">=2.6.20",
-        "sabre/uri" : ">=1.0,<3.0.0"
+        "sabre/uri" : ">=1.0,<1.3"
     },
     "authors" : [
         {


staabm:
looks like a error in sabre/uri then. composer should automatically pick the right version based on the php-requirement.

this fix is a workaround.

marknl:
Hmm, I've did some more testing after reading these comments and my assessment was incorrect. It works just fine.

evert:
Cool ;) if you do ever want to make sure you're getting 1.x, even on PHP 7.x, the easiest is to make sure your own package depends on sabre/uri 1.x itself

PR #116, opened 2017-01-10
set correct return type

lukassteiner:
This function returns a string, but the PHPDoc comment currently says it returns void.


evert:
Thank you!

PR #118, opened 2017-01-11
removed scalar strict type

staabm:
this method supports more types then `string` (as per phpdoc)

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -175,7 +175,7 @@ class Service {
      * @param string|array|XmlSerializable $value
      * @return string
      */
-    function write(string $rootElementName, $value, string $contextUri = null) {
+    function write($rootElementName, $value, string $contextUri = null) {
 
         $w = $this->getWriter();
         $w->openMemory();


evert:
I do think it should be a string in this case, even if the api might support non-strings. The argument is meant to be the rootElementName, and nothing else. I'll drop into the other issue to give an alternative.

staabm:
@evert ok then we should change the phpdoc...

evert:
I think the phpdoc is mostly correct. I removed all @param lines from phpdocs where they are redundant with the actual typehint. Or is the first paragraph maybe confusing

staabm:
your file is using strict types. when the caller of your file also uses strict-types you can only pass strings into this method, no matter what the phpdoc additionally defines.

evert:
Yes but there is no phpdoc for that parameter. There is only the type hint.

staabm:
Ohh.. i see... there is only one phpdoc for the middle param. Everything fine then.

Issue #107, opened 2016-09-12
mixed text/element in node discards text


admonkey:
If my document has mixed text/elements in a node, for example:

```
<root>
   <node>
      some text
      <sub>
          deep text
      </sub>
   </node>
</root>
```

then how can I parse `some text`?

It seems to be simply ignored/discarded when parsed.

Thanks!

evert:
Hi @admonkey,

This is by design. The goal of the default deserializer is to easily parse the most common-case. Adding support for mixing text and elements would make parsing unnecessarily complicated.

However, there are specific areas where this makes sense. One example is XHTML embedded into Atom files.

So the question here is, what do you want to do with the data? If your actual use-cases is indeed XHTML inside Atom, you likely want access to the entire sub-document. In that case you'll probably want:

    $service->elementMap('node', 'Sabre\Xml\Elements\XmlFragment');

If there's another more specific parsing strategy you want, you just have to define your own deserializer:

    $service->elementMap('node', function(Sabre\Xml\Reader $reader) {

        // do stuff here and then return.
    });

evert:
Closing this issue, but feel free to follow up and ask more questions.

PR #119, opened 2017-01-27
added a mixedContent deserializer

### LINKED ISSUE #112: MixedContent Support
In some document one needs to support mixed-content like

```xml
<p>This is some text <extref href="..">and a inline tag</extref>and even more text</p>
```

in this case content is not only between tags, but tags are also mixed together. This form of content is not possible with the current deserializers.

for this case we build this small deserializer function:

```php
    $mixedContent = function(Sabre\Xml\Reader $reader) {

        // If there's no children, we don't do anything.
        if ($reader->isEmptyElement) {
            $reader->next();
            return [];
        }

        $previousDepth = $reader->depth;

        $content = array();
        $reader->read();
        while (true) {
            if ($reader->nodeType == XMLReader::ELEMENT) {
                $content[] = $reader->parseCurrentElement();
            } else if ($reader->depth >= $previousDepth && in_array($reader->nodeType, [XMLReader::TEXT, XMLReader::CDATA, XMLReader::WHITESPACE])) {
                $content[] = $reader->value;
                $reader->read();
            } else if ($reader->nodeType == XMLReader::END_ELEMENT) {
                // Ensuring we are moving the cursor after the end element.
                $reader->read();
                break;
            } else {
                $reader->read();
            }
        }
        return $content;
    };
```

when time allows I will open a PR and provide some tests for it. If anyone beats me todo it, feel free to pick the code up and provide some tests.

staabm:
Closes #112. Refs #107.

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -249,13 +249,17 @@ function repeatingElements(Reader $reader, string $childElementName) : array {
 /**
  * This deserializer helps you to deserialize structures which contain mixed content like this:
  * 
- * <p>some text <extref href="..">and a inline tag</extref>and even more text</p>
+ * <p>some text <extref>and a inline tag</extref>and even more text</p>
  *
  * The above example will return
  *
  * [
  *     'some text',
- *     'and a inline tag',
+ *     [
+ *         'name'       => '{}extref',
+ *         'value'      => 'and a inline tag',
+ *         'attributes' => []
+ *     ],
  *     'and even more text'
  * ]
  *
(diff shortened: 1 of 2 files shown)


lib/Deserializer/functions.php
```diff
@@ -245,3 +245,49 @@ function repeatingElements(Reader $reader, string $childElementName) : array {
...
+ * 
+ * <p>some text <extref href="..">and a inline tag</extref>and even more text</p>
+ *
+ * The above example will return
+ *
+ * [
+ *     'some text',
+ *     'and a inline tag',
+ *     'and even more text'
+ * ]
```
evert: I  don't think this documentation is correct. It would  also contain elements, no?

staabm: Ohh missed that

staabm: fixed.

PR #121, opened 2017-03-03
Fix default namespace attributes

praalhans:
Currently using namespaceMap with empty key does correctly write elements, but does not correctly write attributes (it writes `:attribute=` instead of `attribute=`).

This should fix the problem.

### DIFF
--- a/lib/Writer.php
+++ b/lib/Writer.php
@@ -248,7 +248,8 @@ class Writer extends XMLWriter {
         if (array_key_exists($namespace, $this->namespaceMap)) {
             // It's an attribute with a namespace we know
             return $this->writeAttribute(
-                $this->namespaceMap[$namespace] . ':' . $localName,
+            	$this->namespaceMap[$namespace] === '' ? $localName :
+                	($this->namespaceMap[$namespace] . ':' . $localName),
                 $value
             );
 


praalhans:
Note that the behaviour is still a weird edge case. The XML specification does not state how one can refer to the default namespace, if it is declared. Consider

```
<root xmlns="http://example.org" xmlns:x="http://example.net/other">
<element attr="lives in default" />
<o:other aaa="lives in o" />
<o:other :back="invalid XML syntax" />
</root>
```

At least with this fix, the library does not produce non-well-formed XML. Further research or contact with the authors of the XML spec is necessary to get more information regarding this edge case.

evert:
I'm a little confused. What XML do you hope to produce, and what PHP code are you using to try and produce that (but failing).

praalhans:
Hi @evert!

Thank you for your reply. What I am trying to do is the following:

```
$service = new \Sabre\Xml\Service();
$service->namespaceMap = [
	Order::NS => '',
];
$service->elementMap = [
	'{' . Order::NS . '}order' => Order::class,
];
```

I wish to set the default namespace of the serialized output document to use the empty prefix. This means, that it should not use any prefix at all. The output thus should become:

```
<?xml version="1.0"?>                                                                                
<order xmlns="http://dropcart.nl/2017/03/DropCart/Contracts/Order" ref="1" status="PAID">            
 <problem type="ProblemDetailCustomer">                                                              
  <message>Missing customer details</message>                                                        
  <description>A delivery address is required for the selected shipping method.</description>        
  <key>delivery_address</key>                                                                        
 </problem>                                                                                          
 <!-- et cetera -->
</order>
```

Observe that the attributes of the root element (`ref` and `status` are always local to the element). However, without this patch, the output looks like this:

```
<?xml version="1.0"?>                                                                                
<order xmlns="http://dropcart.nl/2017/03/DropCart/Contracts/Order" :ref="1" :status="PAID">            
```

The code I use to construct this document is the following:

```
$service->write('{' . Order::NS . '}order', $object);
```

Where `Order` is an object implementing `XmlSerializable`. Within the serialization I use a customized ValueObject-esque deserialization mechanism. So the class has a definition akin to:

```
class Order extends \DropCart\Contracts\DataType {
	
	const NS = "http://dropcart.nl/2017/03/DropCart/Contracts/Order";
	
	public $ref;
	public $status;
...
	
	public function xmlSerialize(\Sabre\Xml\Writer $writer) {
		\DropCart\Contracts\XML::serialize($writer, $this, self::NS, ['ref', 'status']);
	}
	
	public static function xmlDeserialize(\Sabre\Xml\Reader $reader) {
		return \DropCart\Contracts\XML::deserialize($reader, self::class, self::NS, ['ref', 'status'], [
				'context' => ProductContext::class,
				'bag' => DetailProduct::class,
				'total' => Price::class,
				'customer' => DetailCustomer::class,
				'payment' => DetailPayment::class,
				'problem' => Problem::class
		]);
	}
	
}
```

where the custom serialization function is adapted from the Sabre library:

```
	static function serialize(Writer $writer, $valueObject, string $namespace, $attributes = []) {
		// Copied from Sabre internal implementation, adapted to also support specifying
		// which variables need be written as attributes to the element. 
		$variables = get_object_vars($valueObject);
		// Note that the variables are listed by get_object_vars in the order from derived class to super class
		// But, for extensibility reasons, XML Schema expects extensions of types always to come after the
		// elements of a base class. Hence, we reverse the array. This means that occurrences of variable declarations
		// in the code are reversed with regard to the output XML.
		$variables = array_reverse($variables);
		
		foreach ($attributes as $key) {
			if (isset($variables[$key])) {
				$writer->writeAttribute('{' . $namespace . '}' . $key, (string)($variables[$key]));
			}
		}
		foreach ($variables as $key => $val) {
			if (in_array($key, $attributes)) continue;
			if (is_array($val)) {
				// If $val is an array, it has a special meaning. We need to
				// generate one child element for each item in $val
				foreach ($val as $child) {
					$writer->writeElement('{' . $namespace . '}' . $key, $child);
				}
			} elseif ($val !== null) {
				$writer->writeElement('{' . $namespace . '}' . $key, $val);
			}
		}
	}
```

As you can see, I have adapted the serializer so to choose which member fields become attributes and which are written as usual (as nested elements). The `writeAttribute` call uses a fully qualified name, in Clark notation. Then, when we map the namespace to `''`, i.e. the default namespace of a document (as can be seen by using `xmlns="..."` in the XML document, the attributes are incorrectly written due to this code:

```
if (array_key_exists($namespace, $this->namespaceMap)) {
	// It's an attribute with a namespace we know
	return $this->writeAttribute($this->namespaceMap[$namespace] . ':' . $localName, $value);
}
```

Here, if the namespace (which I specify using Clark notation for the attribute) is mapped, it incorrectly does not see that the map keys has an in-bound special value of `''` that signals the default namespace. Hence, it incorrectly outputs `:attrName="some...value"` instead of `attrName`.

I did not find anything in XML specifications to explicitly refer to attributes within the default namespace, i.e. the following document is problematic:

```
<?xml ... ?>
<root xmlns="urn:default" xmlns:e="urn:extension">
<element attrib="value" /> <!-- element in default NS -->
<element e:attrib="value" /> <!-- attrib in extension NS -->
<e:element :attrib="value" /> <!-- element in extension NS, attrib in default NS? WRONG -->
</root>
```

A solution to this last example would be to force, whenever refering to an attribute in the default namespace within an element of a non-default namespace, to add another namespace prefix:

```
<?xml ... ?>
<root xmlns="urn:default" xmlns:b="urn:default" xmlns:e="urn:extension">
<e:element b:attrib="value" /> <!-- element in extension NS, attrib in default NS (via prefix b) -->
</root>
```

The code simply cannot detect these subtleties as it is now: to handle this edge case, one has to keep track of the current namespace of the current written element, check whether the attribute written is in the same namespace (then simply drop the prefix) or whether the attribute is in a different namespace. If the attribute is in a different namespace, either check whether the namespace has a well-defined prefix (i.e. not empty) and use that prefix to write the attribute, or if it refers to the default namespace (the empty prefix) then copy over the default prefix to a fresh prefix name with the same namespace and use that fresh prefix to refer to that attribute.

To prevent all this complexity in the implementation, this patch worksaround this problem completely by just dropping the colon if the attribute lives in the default namespace. Clearly not all cases are now correct (and even incorrect, e.g. if a base attribute is written to an element not living in the default namespace, as it gets serialized with a local attribute name that is incorrect). However, currently the implementation produces invalid XML files, which is even a lesser solution.

Hope this helps.

evert:
Hi!

The previous build failed for 2 reasons:

1. We had some problems with PHPUnit changing their API. I fixed this now.
2. There's some linting problems in your PR.

If you fix the linting issues, I will merge your change. Thanks for the detailed explanation, and I'm sorry for the slow response.

staabm:
I will close this PR because of inactivitiy. feel free to re-open it, in case it gets back to live

Issue #92, opened 2016-03-14
Service: missing encoding


staabm:
We produce XML for consumption in glassfish and the XMLs cannot be consumed because of missing encoding declarations.

Hacking `Service::write` with `$w->startDocument('1.0', 'UTF-8')` did the trick for us.
How should the lib consumer tell sabre-xml which encoding the generated XMLs use?

https://github.com/fruux/sabre-xml/blob/master/lib/Service.php#L186

evert:
Well, the default for XML is utf-8 as far as I know. So I feel obligated to ask: is this a known glassfish bug?

staabm:
ok, just had another talk with my colleague which reported the issue.

the actual issue is, that `$w->startDocument()` renders `<?xml version="1.0"?>` but with `$w->startDocument('1.0', 'UTF-8')` it renders `<?xml version="1.0" encoding="UTF-8"?>` and the `encoding` attribute is required by glassfish

evert:
Hey,

So, I understand the issue. I guess what I'm saying is: this should be treated as a glassfish bug, because specifying `encoding` is not required for XML, and glassfish should assume it's UTF-8.

So I am curious if this is a known Glassfish bug. Is it reported somewhere? Is it maybe fixed in a later version?

evert:
If it turns out this is not fixable on Glassfish's end, we can definitely come up with a workaround. But I feel that it's important to first address the bug where it actually lives ;) And right now that seems to be glassfish.

staabm:
Agree. Where did you find that its not fixable in glassfish?

Does your last comment mean the workaround will land in sabre-xml?

staabm:
I misread the comment. I will check the glassfish bugtracker

evert:
I'm closing this ticket because it's been a while since the last comment.

If this is indeed still an issue, feel free to comment here so we can continue discussing.

davidmpaz:
Hi,
first of all, a big thanks to you @evert for sharing your work and taking time to support us. It is really demanding.

Regarding this issue I would like, if possible, to continue the discussion. I took a read to this thread and tried to check whether client problems arise, wanted to understand also where the problem lies. I dont know if I missed something already so I will describe here. Sorry if post goes long.

[...]

PR #122, opened 2017-03-15
Add possibility to specify encoding in Service

davidmpaz:
Allow for setting up the encoding of the writer in Service class

Taking bit more of initiative :) as proposed here #92. This is the small change that helped me with encoding.

Thanks!


evert:
I don't want this feature for reasons stated in #42.

PR #128, opened 2017-07-18
[master] Fix invalid PHPDocs

LukasReschke:
To make https://github.com/etsy/phan happy:

```
apps/dav/lib/CardDAV/SyncService.php:255 PhanNonClassMethodCall Call to method getResponses on non-class type void
apps/dav/lib/CardDAV/SyncService.php:259 PhanNonClassMethodCall Call to method getSyncToken on non-class type void
```

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -131,7 +131,9 @@ class Service {
      *
      * @param string|string[] $rootElementName
      * @param string|resource $input
-     * @return void
+     * @param string|null $contextUri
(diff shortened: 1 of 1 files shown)


LorbusChris:
Can we get a new release with this patch?

staabm:
sure, will take care of it later this week.

staabm:
here we go https://github.com/sabre-io/xml/releases/tag/2.1.0

PR #132, opened 2017-08-17
Prevent infinite loops for empty props element

rullzer:
If you do a propfind with:

```xml
<?xml version="1.0"?>
<a:propfind xmlns:a="DAV:">
  <a:prop></a:prop>
</a:propfind>
```

And have this parsed:

```php
use Sabre\DAV\Xml\Service;

$content = '<?xml version="1.0"?><a:propfind xmlns:a="DAV:"><a:prop></a:prop></a:propfind>';

$xml = new Service();
$xml->expect('{DAV:}propfind', $content);
```

Then this results in an infinite loop.
This patch breaks out of the loop once the reader fails to read.

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -74,7 +74,9 @@ function keyValue(Reader $reader, string $namespace = null) : array {
                 $values[$clark] = $reader->parseCurrentElement()['value'];
             }
         } else {
-            $reader->read();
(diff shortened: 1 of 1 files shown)


LorbusChris:
@rullzer any progress on this?

rullzer:
It is still dangling somewhere on my todo... But that todo is long

PR #133, opened 2017-09-06
[1.5] Correct Service->mapValueObject() PHPDoc

endelwar:


staabm:
thx for your PR. I will close it now, since we no longer have those typehints on our master branch and fixing it only in the old series doesn't provide enough value.

sorry for taking so long to get to this PR. I didn't realize it.

PR #144, opened 2018-09-21
Enriched VOs

vsouz4:
We need real VOs when trying to map VOs to/from XML, meaning we need:

1. ability to use final classes
2. ability to use private or protected properties
3. define a named constructor (aka factory method)
4. work with getters to retrieve the values
5. accept arrays of VOs

Right now we can only rely on classes with public properties and array support for VOs is weird/broken. I've added an optional parameter when adding a VO (`Service::mapValueObject`) which is the factory method name. If this is set then we go through the "new flow" when reading/writing XMLs with VOs. This means these changes are backward compatibles ("old" unit tests are still working at least), so only people who use this optional parameter will get the results of this merge request.

I've added a unit test with an example of XML and VOs that the library should start to support.


evert:
Hi @vsouz4 ,

All of this can be implemented as a user of the library, without modifying the library. `mapValueObject` really is a feature intended to address the simplest cases, you are squarely in a situation where `mapValueObject` is too simplistic, so my answer to that is to simply not use it.

`mapValueObject` is a small abstraction on top of `elementMap` and `classMap`, so your version of Value Objects should also just use those 2 primitives directly, instead of extending the meaning of `mapValueObject`.

Your specific use-case for value objects includes assuming that there are getters/setters and that a value object has a static factory method that instantiates itself. I don't think there's anything wrong with that pattern, but I am not sure if this is a good 'default pattern' for this library to support out of the box. Especially since this library already has a few default patterns and it might end up adding a bunch of cruft that further confuses users (I found that users are already a bit confused about the best approach to using this library).

I'm also of the opinion that re-using `mapValueObject` to support a second pattern is worse than adding new functions specifically for this case. It violates the single responsibility pretty hard.

So while I really appreciate the effort put in this feature, I don't think it's a good fit for this library.

vsouz4:
There's only 1 main rule that bases everything I've done in this merge request: value objects should be immutable 
 (https://martinfowler.com/bliki/ValueObject.html, https://en.wikipedia.org/wiki/Value_object)

1) This means a PHP class with public properties is not a value object. A PHP class defined as final plus private properties is more adequate. 
2) If it's immutable you need to receive all arguments in the constructor (meaning no setters).
3) If you have private properties, you need getters.
4) If you have so far a final class with private properties, getters and a constructor type hinting the arguments with the correct types and you need to build them from a XML representation, you need a constructor that supports it (if this was java we would overload the constructor) but PHP relies on factory methods (named constructors) for that (http://verraes.net/2014/06/named-constructors-in-php/, https://martinfowler.com/articles/refactoring-document-load.html, https://stackoverflow.com/a/9442908).

The only choice I made is that I'm calling the named constructor with the arguments as string 'cause in the XML those values are string. `fromXml` could receive the XML element itself and it would be up to the vo using one your functions to read it but I'm not sure this would be the best approach, since we can already extract the info the VO needs and use the named constructor with all arguments as string (as I'm currently doing).

Regarding using the same method in service class `mapValueObject`, that's my understanding for having a organic change from the old VO style this library supports to the more adequate VO representation. There should be no support for the old VO style anymore since that's just not a VO but this would only come with a major version update (since it breaks backward compatible, would need depreciation notes etc.), so creating one different method for "my version of vo" as you say would imply imho that both can be used and you're fine writing VOs with public properties. This, however, is something I can change, if you prefer to have a new method instead of using the current one.

Completely agree all of these can be implemented in the user library, but then what's the point of open source? Using VOs with public properties is just not reasonable imo and I want to contribute with this library so that it starts supporting a more adequate VO representation (actually the only one logically based on the assumption VOs are immutable).

Can we get to an agreement about this or you really want to keep supporting only php classes with public properties?

evert:
Hi @vsouz4 ,

You're probably right that the use of the term Value Object perhaps wasn't the best. It also seems like a pretty good pattern.

However, I feel that the library already supports this pattern just fine with its existing features. I like the idea of a small 'core'.

The one piece of this that I think could be a good addition in the library is the deserializer function. I like the idea of having a deserializer that can receive a (factory-?) function, reflect its arguments and call the function mapping child elements, but it needs a new name.

Maybe just something like

    function functionCaller(Reader $reader, callable $func, $namespace)

I think this is a nice generic useful building block. I don't love `functionCaller` but it's all I could think of right now :P

But yea, I don't feel so good about the serializer (inspecting private members) or the changes to the Service class.

vsouz4:
Cool, I've reseted my branch with your master and reapplied only the Deserializer changes in a separate function as you suggested. As you said, it's more generic this way and whoever wants to work with it like I want/need (with VOs) or with a callable that does anything else, it works. I've added unit tests with all types of callable php accepts (string, array, closure etc.) and everything is working fine. I could run my example with VOs by defining the elementMaps manually but I'm ok with it since this seems to be the usual way to work with this library, so no need to change the Service class or anything else ;)

What do you think now?

lib/Deserializer/functions.php
```diff
@@ -295,3 +295,42 @@ function mixedContent(Reader $reader) : array {
...
+
+/**
+ * The functionCaller deserializer turns an xml element into whatever your callable return.
+ *
+ * You can use, e.g., a named constructor (factory method) to create an object using
+ * this function.
+ *
+ * @return mixed
+ */
+function functionCaller(Reader $reader, callable $func, string $namespace) {
```
staabm: Maybe $factory instead of $func

vsouz4: This was @evert suggestion actually. In fact, this callable can be used as a factory, but this function is more generic than that. We could use this callable to return an array (or anything else), and imo $factory is more related to objects/classes.

vsouz4:
Hi @evert, we want to use the new deserializer function in our project :) I've made the fixes u suggested and also @staabm 's as well. What do you think? Can we merge it?

vsouz4:
Fixed CS and code coverage!

PR #145, opened 2018-09-24
Test with php7.3

DeepDiver1975:


DeepDiver1975:
seems to work with php 7.3 out of the box

staabm:
Thx

PR #149, opened 2018-10-09
Properly detect xml parse errors in any case

DeepDiver1975:
This was detected while running litmus with sabre dav 4 ....


lib/Reader.php
```diff
@@ -70,6 +70,12 @@ function parse() : array {
             }
             $result = $this->parseCurrentElement();
 
+            // last line of defense in case errors did occur above
```
staabm: just to get it right: we check for a error happing in `$this->parseCurrentElement();` with this new case?

tests/Sabre/Xml/ServiceTest.php
```diff
@@ -123,6 +125,25 @@ function testExpect() {
...
+        $util->elementMap = [
+            '{DAV:}propfind'       => PropFindTestAsset::class,
+        ];
+        $util->namespaceMap = [
+            'http://sabre.io/ns' => 's',
+        ];
+        $result = $util->expect('{DAV:}propfind', $xml);
+
+    }
+
```
staabm: I guess this needs CS fixes

lib/Reader.php
DeepDiver1975: that's actually true - shall we move it inside of parseCurrentElement()?

staabm: no we can't as parseCurrentElement() is not expected to throw exceptions.

DeepDiver1975:
time to release ....

PR #158, opened 2019-01-08
Prevent infinite loop on empty xml elements

DeepDiver1975:


staabm:
after getting in the changes discussed in https://github.com/sabre-io/xml/pull/159 we get move ahead.

thx

rullzer:
:+1: good stuff @DeepDiver1975

PR #159, opened 2019-01-09
[1.5] Prevent infinite loop on empty xml elements

DeepDiver1975:
Brings #158 to the 1.5 branch

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -66,9 +66,20 @@ function keyValue(Reader $reader, $namespace = null) {
         return [];
     }
 
+    if (!$reader->read()) {
+        $reader->next();
+
+        return [];
+    }
+
+    if (Reader::END_ELEMENT === $reader->nodeType) {
+        $reader->next();
+
+        return [];
+    }
+
     $values = [];
 
-    $reader->read();
     do {
 
         if ($reader->nodeType === Reader::ELEMENT) {
@@ -79,7 +90,9 @@ function keyValue(Reader $reader, $namespace = null) {
                 $values[$clark] = $reader->parseCurrentElement()['value'];
             }
         } else {
-            $reader->read();
+            if (!$reader->read()) {
+                break;
+            }
         }
     } while ($reader->nodeType !== Reader::END_ELEMENT);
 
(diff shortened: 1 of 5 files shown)


lib/Deserializer/functions.php
```diff
@@ -79,11 +90,15 @@ function keyValue(Reader $reader, $namespace = null) {
...
         } else {
-            $reader->read();
+            if (!$reader->read()) {
+                break;
+            }
         }
     } while ($reader->nodeType !== Reader::END_ELEMENT);
 
-    $reader->read();
+    if (!$reader->read()) {
```
froschdesign: Why was the `if` statement added here? This duplicates also the return of this method.

DeepDiver1975: this is true - thx for having a closer look. 

I added this for debugging purpose to see if read fails at this point under any condition .....

tests/Sabre/Xml/ServiceTest.php
```diff
@@ -123,6 +125,40 @@ function testExpect() {
         );
     }
 
+    /**
+     * @expectedException \Sabre\Xml\LibXMLException
```
froschdesign: The support for this annotation will be removed with version 9 of PHPUnit. This would not be a problem if this package did not use the "death star version constraint" on PHPUnit.

* https://github.com/sebastianbergmann/phpunit/issues/3333
* https://thephp.cc/news/2017/02/death-star-version-constraint

My suggestion: use the [`expectException`](https://phpunit.readthedocs.io/en/7.4/writing-tests-for-phpunit.html#testing-exceptions) method instead and define a version for PHPUnit in the Composer configuration – version 8 is around the corner.

staabm: at first we should get rid of the "death star version constraint".

changes to our unit tests itself are welcome via separate pull requets

froschdesign: @staabm 
> thx for noticing.

Right, this was only a notice and should be done in a separate PR.

DeepDiver1975: This is the 1.5 branch - it is on pure maintenance mode. No idea if we should put any further man power into this.

staabm: > This is the 1.5 branch - it is on pure maintenance mode. No idea if we should put any further man power into this.

the mentioned issues should be fixed on master.

in 1.5 I would only fix the "death star version constraint"

staabm:
feel free to merge, in case this is ready.

DeepDiver1975:
> feel free to merge, in case this is ready.

I'm preparing the release right in here as well ... then merge .... then repeat on all other PRs ....

DeepDiver1975:
https://github.com/sabre-io/xml/releases/tag/1.5.1

PR #160, opened 2019-01-09
[2.0] Prevent infinite loop on empty xml elements

DeepDiver1975:
port of #158 to 2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ChangeLog
 =========
 
+2.0.2 (2019-01-09)
+------------------
+
+* #161: Prevent infinite loop on empty xml elements
+
+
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
https://github.com/sabre-io/xml/releases/tag/2.0.2

evert:
I don't think you have to maintain the 2.0 branch. afaik 2.1 is 100% compatible, and unlike the `sabre/dav` package, every other package does use semantic versioning.

DeepDiver1975:
> I don't think you have to maintain the 2.0 branch. afaik 2.1 is 100% compatible, and unlike the `sabre/dav` package, every other package does use semantic versioning.

Didn't hurt that much to release this ;-)

PR #161, opened 2019-01-09
[2.1] Prevent infinite loop on empty xml elements

DeepDiver1975:
port of #158 to 2.1

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -61,9 +61,20 @@ function keyValue(Reader $reader, string $namespace = null) : array {
         return [];
     }
 
+    if (!$reader->read()) {
+        $reader->next();
+
+        return [];
+    }
+
+    if (Reader::END_ELEMENT === $reader->nodeType) {
+        $reader->next();
+
+        return [];
+    }
+
     $values = [];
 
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
https://github.com/sabre-io/xml/releases/tag/2.1.2

lib/Deserializer/functions.php
```diff
@@ -137,7 +150,17 @@ function enum(Reader $reader, string $namespace = null) : array {
...
+        $reader->next();
+
+        return [];
+    }
+
+    if (Reader::END_ELEMENT === $reader->nodeType) {
+        $reader->next();
+
+        return [];
+    }
```
evert: This feels a bit hacky. Makes me wonder if this is there is a cleaner way to do this. Would for instance just a single `$reader->next()` do the job (removing `$reader->read()`?)

DeepDiver1975: THX for your feedback @evert 

I can have a look at this .... 
Since I already tagged the versions are you okay in accepting this implementation as is?

evert: Yea np. If it works it works. It was a mild concern. The tests pass which is the most important =)

PR #169, opened 2019-08-14
Throw exception when empty inputs found

DeepDiver1975:
When input retrieved from stream_get_contents
is empty string, then it would be safe to throw
exceptions from the parse and expect methods.

Signed-off-by: Sujith H <sharidasan@owncloud.com>

master version of #166

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -115,6 +115,11 @@ class Service
             // Unfortunately the XMLReader doesn't support streams. When it
             // does, we can optimize this.
             $input = (string) stream_get_contents($input);
+
+            // If input is an empty string, then its safe to throw exception
+            if ('' === $input) {
+                throw new ParseException('The input element to parse is empty. Do not attempt to parse');
+            }
         }
         $r = $this->getReader();
         $r->contextUri = $contextUri;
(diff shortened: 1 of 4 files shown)


DeepDiver1975:
@staabm mind reviewing? THX

lib/Service.php
```diff
@@ -115,6 +115,11 @@ public function parse($input, string $contextUri = null, string &$rootElementNam
             // Unfortunately the XMLReader doesn't support streams. When it
             // does, we can optimize this.
             $input = (string) stream_get_contents($input);
+
+            // If input is an empty string, then its safe to throw exception
+            if ('' === $input) {
+                throw new ParseException('The input element to parse is empty. Do not attempt to parse');
```
staabm: not sure I get the meaning of this error message? what do you try to tell the caller?

lib/Service.php
```diff
@@ -115,6 +115,11 @@ public function parse($input, string $contextUri = null, string &$rootElementNam
             // Unfortunately the XMLReader doesn't support streams. When it
             // does, we can optimize this.
             $input = (string) stream_get_contents($input);
+
+            // If input is an empty string, then its safe to throw exception
```
staabm: we should mention that we get into an infinite loop when we don't throw

DeepDiver1975:
FYI - this is just the merge of the original PR which targeted 2.1 - see #166

PR #173, opened 2020-01-29
Update dependencies and tooling

phil-davis:
- drop PHP 7.0
- move to phpunit 7
- use later phpstan
- run php-cs-fixer in CI

### DIFF
--- a/lib/Reader.php
+++ b/lib/Reader.php
@@ -147,7 +147,9 @@ class Reader extends XMLReader
                 throw new ParseException('This should never happen (famous last words)');
             }
 
-            while (true) {
+            $keepOnParsing = true;
+
+            while ($keepOnParsing) {
                 if (!$this->isValid()) {
                     $errors = libxml_get_errors();
 
@@ -169,7 +171,8 @@ class Reader extends XMLReader
                     case self::END_ELEMENT:
                         // Ensuring we are moving the cursor after the end element.
                         $this->read();
-                        break 2;
+                        $keepOnParsing = false;
+                        break;
                     case self::NONE:
                         throw new ParseException('We hit the end of the document prematurely. This likely means that some parser "eats" too many elements. Do not attempt to continue parsing.');
                     default:
--- a/lib/Serializer/functions.php
+++ b/lib/Serializer/functions.php
@@ -164,8 +164,6 @@ function standardSerializer(Writer $writer, $value)
     } elseif (is_callable($value)) {
         // A callback
         $value($writer);
-    } elseif (is_null($value)) {
-        // nothing!
     } elseif (is_array($value) && array_key_exists('name', $value)) {
         // if the array had a 'name' element, we assume that this array
         // describes a 'name' and optionally 'attributes' and 'value'.
@@ -204,7 +202,7 @@ function standardSerializer(Writer $writer, $value)
         }
     } elseif (is_object($value)) {
         throw new InvalidArgumentException('The writer cannot serialize objects of class: '.get_class($value));
-    } else {
+    } elseif (!is_null($value)) {
         throw new InvalidArgumentException('The writer cannot serialize values of type: '.gettype($value));
     }
 }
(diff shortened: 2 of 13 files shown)


phil-davis:
The newer version of phpstan reports:
```
------ ------------------------------------------------------- 

  Line   Reader.php                                             

 ------ ------------------------------------------------------- 

  187    Unreachable statement - code above always terminates.  

 ------ ------------------------------------------------------- 

 ------ ---------------------------------------------------------------------- 

  Line   Serializer/functions.php                                              

 ------ ---------------------------------------------------------------------- 

  167    Call to function is_null() with array|object will always evaluate to  

         false.                                                                

         💡 Because the type is coming from a PHPDoc, you can turn off this     

         check by setting treatPhpDocTypesAsCertain: false in your             

         /home/travis/build/sabre-io/xml/phpstan.neon.                         

 ------ ---------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------- 

  Line   Writer.php                                                       

 ------ ----------------------------------------------------------------- 

  130    Parameter #3 $uri of method XMLWriter::startElementNs() expects  

         string, null given.                                              

 ------ ----------------------------------------------------------------- 

                                                                                

 [ERROR] Found 3 errors                                                         
```

I am looking at those.

lib/Reader.php
```diff
@@ -169,7 +171,8 @@ public function parseInnerTree(array $elementMap = null)
                     case self::END_ELEMENT:
                         // Ensuring we are moving the cursor after the end element.
                         $this->read();
-                        break 2;
```
phil-davis: This `break 2` was causing phpstan to report:
```
Line 187    Unreachable statement - code above always terminates.
```
phpstan seems to not realise that `break 2` is a way out of both the `switch` and `while` statements.

It is easy enough to refactor the code to have a boolean to control the `while` loop. And maybe that makes it easier to read?

lib/Serializer/functions.php
```diff
@@ -165,8 +164,6 @@ function standardSerializer(Writer $writer, $value)
     } elseif (is_callable($value)) {
         // A callback
         $value($writer);
-    } elseif (is_null($value)) {
```
phil-davis: phpstan complains:
```
167    Call to function is_null() with array|object will always evaluate to false
```
It knows that `$value` at this point might be an array or object.

After looking at the code, there are `elseif` cases below that explicitly handle the `array` and `object` cases. So I refactored so that `is_null()` is not needed here.

phpstan.neon
```diff
@@ -1,3 +1,6 @@
 parameters:
   level: 5
-  bootstrap: %currentWorkingDirectory%/vendor/autoload.php
+  ignoreErrors:
+    -
+      message: '!Parameter #3 \$uri of method XMLWriter::startElementNs\(\) expects string, null given.!'
```
phil-davis: phpstan complains:
```
130    Parameter #3 $uri of method XMLWriter::startElementNs() expects string, null given.
```
The code is:
```
                $result = $this->startElementNS(
                    '' === $this->namespaceMap[$namespace] ? null : $this->namespaceMap[$namespace],
                    $localName,
                    null
                );
```
Actually phpstan is correct - `startElementNS` is supposed to take a string and not allow `null` to be passed.

But also I found https://www.php.net/manual/en/function.xmlwriter-start-element-ns.php#76920
```
If you don't want any namespace prefix at all but still want the xmlns attribute, set $prefix to null.
```

So I ignored this phpstan error, rather than "fixing" the code.

lib/Reader.php
staabm: we should report a upstream issue.. in case you dont want/don't have time to do it, i will do it.

just let me know.

phil-davis: @staabm you know better exactly about upstream. Please report.

staabm: will do. thx.

IMO the newly introduced logic is better to read, because its now obvious what this single spot is trying to achieve. the `break 2` could also be read as a mistake

staabm: lets see: https://github.com/phpstan/phpstan/issues/2926

PR #174, opened 2020-01-29
Use phpunit8

phil-davis:
See last commit.

This is on top of #173

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ composer.lock
 
 # Tests
 tests/cov
+tests/.phpunit.result.cache
 .*.swp
 
 # Composer binaries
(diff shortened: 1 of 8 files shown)


staabm:
> This is on top of #173

hmm .. I guess I merged it in the wrong order.. could you rebase the PR?

how was I supposed to handle this situation instead?

phil-davis:
> how was I supposed to handle this situation instead?

I have no idea why it said that there was a conflict. I rebased and did not have to manually resolve any conflict, and the diffs look fine.

phil-davis:
@staabm IMO all the "sub"-repos now have PHP 7.4 support and run phpunit8.

So whenever you like, releases can be made. And then someone can update `sabre-io/dav` with all the dependencies and make it support PHP 7.4 and phpunit8.

Then that will let downstream users of these packages use the new versions and officially support PHP 7.4

staabm:
Thank you so much - great job. Well done

PR #177, opened 2020-02-04
Do not --prefer-source for composer update

phil-davis:
Like https://github.com/sabre-io/skel/pull/13


phil-davis:
Previously `phpunit`  ran 109 tests. 
Now it still runs 109 tests - good.

PR #179, opened 2020-02-10
Use phpunit 9 where possible

phil-davis:
https://phpunit.de/announcements/phpunit-9.html

2nd commit: In Travis, only bother to report/upload coverage if the job actually runs unit tests with coverage enabled.


phil-davis:
`codecov/project` has now decided to find `lib/Version.php` and `lib/Service.php` as "new" files to it. That has just a `const` in it and has no unit test coverage. That is why the codecov is lower.

PR #182, opened 2020-03-30
no empty braces when empty or null the namespaceURI

leli1337:


staabm:
could you elaborate a bit what you are trying to achieve with this change and what the problems looks like you are trying to fix?

leli1337:
@staabm 
Consider an xml file called  **somexml.xml** with this content:

```XML
<?xml version="1.0"?>
<books>
    <book>
        <title>Snow Crash</title>
        <author>Neil Stephenson</author>
    </book>
    <book>
        <title>Dune</title>
        <author>Frank Herbert</author>
    </book>
</books>
```

If you parse it and then convert to XML again (the following code) will result in an xml different from the initial.

```PHP
$path = 'somexml.xml';
$fileStream = fopen($path, 'rw');
$xmlService = new Service;
$result = $xmlService->parse($fileStream);
$xml = $xmlService->write('root', $result);
```

```XML
<?xml version="1.0"?>
<root>
    <book xmlns="">
        <title xmlns="">Snow Crash</title>
        <author xmlns="">Neil Stephenson</author>
    </book>
    <book xmlns="">
        <title xmlns="">Dune</title>
        <author xmlns="">Frank Herbert</author>
    </book>
</root>
```

lib/Reader.php
```diff
@@ -39,6 +39,10 @@ public function getClark()
             return null;
         }
 
+        if (is_null($this->namespaceURI) || empty($this->namespaceURI)) {
```
staabm: I guess this can be simplified to
```suggestion
        if (empty($this->namespaceURI)) {
```

staabm:
could you turn you excellent example into a unit test?

leli1337:
> could you turn you excellent example into a unit test?

Yes, I will create a new unit test for Reader.

staabm:
Per method comment in https://github.com/sabre-io/xml/pull/182/files#diff-3a8e508b4baf36a62dd9aca54f4c6387R30 I guess the fix is at the wrong spot

evert:
Yes, the getClarkNotation shouldn't change. The change should be in the writer, not the reader.

leli1337:
@staabm 
The problem is not in writer. Consider the following code 

```PHP
$array = [
    'name' => 'test',
    'value' => [
        [
            'name' => '{}empty-namespace',
            'value' => 'value',
        ], [
            'name' => 'no-namespace',
            'value' => 'value',
        ],
    ],
];

$xmlService = new Service;
$xml = $xmlService->write('root', $array);
```

When using **{}** it will output empty namespace and if you do not put it it will not output namespace.

The result xml of the code is:
```XML
<?xml version="1.0"?>
<root>
    <test>
        <empty-namespace xmlns="">value</empty-namespace>
        <no-namespace>value</no-namespace>
    </test>
</root>
```

evert:
Yes, but that is a design decision.

Removing that can create collisions with arrays that contain a `value`, `name`, and `attribute` key. It was important to have a canonical non-conflicting value for these, and it was a design goal to make 'reading' and 'writing' the result to yield the same XML document (semantically, not byte-for-byte).

Consider the following:

```php
[
  'name' => 'foo',
  'value' => 'bar',
  'attributes' => ['a' => 'b']
}
```

This could be interpreted as:

```xml
<foo a="b">bar</foo>
```

or:

```xml
<name>foo</name>
<value>bar</value>
<attributes><a>b</a></attributes>
```

By decoding non-namespaced elements to `{}name`, there's no ambiguity.

I realize that it is a little ugly for non-namespaced XML, but deserializers help with this. They basically give you a hook into every element and output exactly the structure you need.

evert:
This library actually ships with 1 such deserializer: KeyValue. It will strip namespaces.

Caveat is that it deserializes strictly to a key value structure, so no more arrays with name, value, attributes. With a good combination of deserializers, you can create a very clean output.

PR #185, opened 2020-05-11
Release 2.2.1

staabm:
closes https://github.com/sabre-io/xml/issues/184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ChangeLog
 =========
 
+2.2.1 (2020-05-11)
+------------------
+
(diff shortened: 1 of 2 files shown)


phil-davis:
@staabm thanks!
I looked in `sabre/dav` and it depends on `sabre/xml` in `composer.json` but does not have a `composer.lock` - so I suppose that anyone that installs `sabre/dav` will now get this patch release of `sabre/xml` automagically. There is no need to touch or release `sabre/dav`
Correct?

staabm:
> There is no need to touch or release `sabre/dav`
> Correct?

everyone invoking `composer update` will get it, right.

in case we get a lot of bug reports because people run into a problem, we could make a new sabre/dav release

PR #190, opened 2020-10-03
Test with PHP8.0

phil-davis:
1) sort out variables in `.travis.yml` for selecting to run php-cs-fixer, phpunit and phpstan
2) Add a matrix entry to run just phpunit on Ubuntu 20.04 "focal" with "nightly" PHP (which is the way to get PHP 8.0 at the moment)
3) When not running php-cs-fixer, explicitly remove it from `composer.json` (because php-cs-fixer does not yet support PHP 8.0, so composer cannot sort out dependencies if we have PHP 8.0 and php-cs-fixer 2.* together)
4) `libxml_disable_entity_loader` is deprecated on PHP 8.0. This is already disabled on PHP 8.0, so this function does not need to be called. Put a conditional check around it. https://php.watch/versions/8.0/libxml_disable_entity_loader-deprecation

Relates to https://github.com/sabre-io/vobject/pull/513

### DIFF
--- a/lib/Reader.php
+++ b/lib/Reader.php
@@ -55,6 +55,7 @@ class Reader extends XMLReader
      */
     public function parse(): array
     {
+        $previousEntityState = null;
         $shouldCallLibxmlDisableEntityLoader = (\PHP_VERSION_ID < 80000);
         if ($shouldCallLibxmlDisableEntityLoader) {
             $previousEntityState = libxml_disable_entity_loader(true);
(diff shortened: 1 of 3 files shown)


lib/Reader.php
```diff
@@ -55,7 +55,10 @@ public function getClark()
      */
     public function parse(): array
     {
-        $previousEntityState = libxml_disable_entity_loader(true);
+        $shouldCallLibxmlDisableEntityLoader = (\PHP_VERSION_ID < 80000);
```
staabm, line 58:
```suggestion
        $previousEntityState = null;
        $shouldCallLibxmlDisableEntityLoader = (\PHP_VERSION_ID < 80000);
```

staabm (approved):
Love it! Thanks

PR #213, opened 2022-07-21
added phpstan types to deserializer functions

staabm:


phpstan-baseline.neon
```diff
@@ -795,6 +780,16 @@ parameters:
...
 
+		-
+			message: "#^Property Sabre\\\\Xml\\\\Service\\:\\:\\$elementMap \\(array\\<string, \\(callable\\(\\)\\: mixed\\)\\|class\\-string\\>\\) does not accept array\\{\\{http\\://sabre\\.io/ns\\}test\\: 'Test\\!'\\}\\.$#"
+			count: 1
+			path: tests/Sabre/Xml/ServiceTest.php
+
+		-
+			message: "#^Property Sabre\\\\Xml\\\\Service\\:\\:\\$namespaceMap \\(array\\<string, class\\-string\\>\\) does not accept array\\<string, string\\>\\.$#"
+			count: 4
+			path: tests/Sabre/Xml/ServiceTest.php
```
staabm, line 791: these test-cases seem to intentionally use test-data which does not adhere to our own api?

phpstan-baseline.neon
```diff
@@ -345,6 +300,36 @@ parameters:
 			count: 1
 			path: tests/Sabre/Xml/Deserializer/FunctionCallerTest.php
 
+		-
+			message: "#^Parameter \\#2 \\$func of function Sabre\\\\Xml\\\\Deserializer\\\\functionCaller expects callable\\(\\)\\: Sabre\\\\Xml\\\\Deserializer\\\\Address, 'Sabre\\\\\\\\Xml…' given\\.$#"
```
staabm, line 304: these tests use some pretty exotic notations for declaring a callable in https://github.com/sabre-io/xml/blob/584d601a7bdec8af7e5de0b18b9b4398249a7bde/tests/Sabre/Xml/Deserializer/FunctionCallerTest.php#L100-L105

I am fine with reporting these as errors, wdyt?

phpstan-baseline.neon
phil-davis: See #216 commit https://github.com/sabre-io/xml/pull/216/commits/517dd4866c1ddd3b77d7e48a84a2c04fb9ec52a3

These can be got rid of by specifying a string that PHPstan recognizes as a class name - I used the built-in `stdClass` 

That makes everything pass. Maybe that is OK to do in the tests. If you think so, then cherry-pick that commit into this PR.

That PR also fixes small things that cs-fixer complains about:
https://github.com/sabre-io/xml/pull/216/commits/631da960f71ae7c4117ac0d90b2d1efebde0e887

phpstan-baseline.neon
phil-davis: Actually `functionCaller` is happy to accept a string that is a class, and anything that is a callable - regardless of the parameters of the callable.

The doc line `@param callable():R $func` in `functionCaller` seems to force PHPstan to only allow passing a callable that is a function with no parameters.

I removed that doc line in commit https://github.com/sabre-io/xml/pull/216/commits/cc4ce0635fb2a84d6ef3900022face9e99114a55 and got rid of these lines of "expected messages" and it passes.

In the tests, locally, I tried passing some string that is not a class/method that exists and that makes phpstan complain. So it is checking that a real callable is being passed.

Maybe cherry-pick that commit?

phil-davis:
@staabm I made a couple of comments above. See what you think...

phpstan-baseline.neon
staabm: picked the commit, thx

phpstan-baseline.neon
staabm: dropping the phpdoc will remove the tooling/type coverage. but I wasn't able to figure out a way, to make phpstan analyse this code properly. so leaving it out for now, as you suggested.

PR #214, opened 2022-07-25
update phpstan 1.x

staabm:


phpstan.neon
```diff
@@ -1,6 +1,9 @@
...
+    - phpstan-baseline.neon
+
 parameters:
-  level: 5
-  ignoreErrors:
-    -
-      message: '!Parameter #3 \$namespace of method XMLWriter::startElementNs\(\) expects string, null given.!'
-      path: lib/Writer.php
+  level: 6
+  phpVersion: 70234 # PHP 7.2.34
```
staabm, line 6: using a fixed php version to make it possible to create a baseline locally on different php versions.

doesn't matter much which version we use here. I am using 7.2 as its on the lower end but 7.1 seemed too old for my feelings

phpstan-baseline.neon
```diff
@@ -0,0 +1,926 @@
...
+			count: 1
+			path: lib/Deserializer/functions.php
+
+		-
+			message: "#^Function Sabre\\\\Xml\\\\Deserializer\\\\repeatingElements\\(\\) return type has no value type specified in iterable type array\\.$#"
+			count: 1
+			path: lib/Deserializer/functions.php
+
+		-
+			message: "#^Negated boolean expression is always false\\.$#"
```
staabm, line 19: Fixed with https://github.com/phpstan/phpstan-src/pull/1551

PR #217, opened 2022-07-26
declare types

staabm:


lib/Service.php
```diff
@@ -252,7 +255,7 @@ public function mapValueObject(string $elementName, string $className)
             return \Sabre\Xml\Deserializer\valueObject($reader, $className, $namespace);
         };
         $this->classMap[$className] = function (Writer $writer, $valueObject) use ($namespace) {
-            return \Sabre\Xml\Serializer\valueObject($writer, $valueObject, $namespace);
```
staabm, line 255: a serializer does not return a vlaue

lib/Writer.php
```diff
@@ -182,6 +184,7 @@ public function startElement($name): bool
      * XMLWriter::startElement doesn't either.
      *
      * @param array|string|object|null $content
+     * @param mixed $content
```
phil-davis, line 187: `$content` is mentioned twice ??

phil-davis (commented):
and cs-fixer is complaining about a few minor things

lib/Writer.php
staabm: good catch

phil-davis (approved):
LGTM

staabm:
thank you phil.

Issue #215, opened 2022-07-25
bump min php versions


staabm:
I think we should strip down the supported php version matrix

what do you guys think about dropping at least 7.1/7.2 ?

see https://www.php.net/supported-versions.php

phil-davis:
IMO just support PHP 7.3 and up. 
7.3 has been out of official support for security patches since late last year (about 7 months ago now). Maybe support for 7.3 can be dropped around the end of this year?

One benefit of dropping the older minor versions of PHP is that gradually more parameter and return types can be used, and new functions relied-on etc.

tacman:
Since 7.3 is at EOL, what about bumping the version and requiring 7.4?  Typed properties in classes are great, among other things.

phil-davis:
> Since 7.3 is at EOL, what about bumping the version and requiring 7.4? Typed properties in classes are great, among other things.

Yes, we are about at that point anyway. Various consumers of these sabre-io repos that I am familiar with have either dropped, or are soon to drop, PHP 7.3 support. 

We could make a branch from current master that can keep the current PHP support and if any interesting/critical/security patches are needed then we can backport patches to that branch. And then in master develop PHP 7.4+ code, put all the possible type declarations in parameters, function returns, variables, removing deprecated stuff... and release as a new major version. That avoids the need to think hard about strict backward-compatibility for methods of classes etc that start having stricter formal interfaces, and for which there could be implementations that extend the class and now have to also add the types.

For the future, some time in 2023, we can discuss dropping PHP 7.4. And that may or may not cause another major version bump, depending on what extra things happened in PHP 8.0+ that are not available in PHP 7.4. But that can be a separate discussion.

staabm:
Agree 👍

phil-davis:
Major version 3 https://github.com/sabre-io/xml/releases/tag/3.0.0 has been released here.
https://github.com/sabre-io/uri has minor release https://github.com/sabre-io/uri/releases/tag/2.3.0 - there was no need to cause a BC-break there, so no need for a major version bump.

I am looking at vobject now. I will leave this issue open, and update as I progress up the tree to finally get sabre/dav sorted out.

DeepDiver1975:
7.4 is min -> close

PR #220, opened 2022-08-12
Set minimum PHP to 7.4

phil-davis:
1) drop support for PHP  7.1 7.2 and 7.3. Support PHP  7.4 8.0 and 8.1
2) bump code style and analysis tools to their latest versions
3) adjust parameter types, return types, constants, variable types.... to be as specific as possible, using features available since PHP 7.4
4) bump the major version from 2 to 3, so that existing systems only grab this new code by specifying a major version bump - because they might have to touch some code (e.g. adding `: void` to a child function...) in order to use this.

This fixes issue #215 for this repo.

There are no other open PRs here, so I think we can do this one right away. If a serious bug is noticed in the old code, and we really need to make a patch release of 2.x then we can make a branch for that and backport the fix... But it is probably not worth bothering to set that up here - it might never be used.


tests/Sabre/Xml/ContextStackTest.php
```diff
@@ -15,11 +15,11 @@
  */
 class ContextStackTest extends TestCase
 {
-    private $stack;
+    private ContextStack $stack;
 
     public function setUp(): void
     {
-        $this->stack = $this->getMockForTrait('Sabre\\Xml\\ContextStackTrait');
+        $this->stack = new ContextStack();
```
phil-davis, line 22: Ref: https://github.com/phpstan/phpstan-phpunit/issues/30

It seems better to use a real class that is just a "wrapper" around the trait, and test that class. This "just works"  for the tests here, they still pass. And it makes it possible to declare the type of `$stack`

lib/Element/Base.php
```diff
@@ -50,7 +52,7 @@ public function __construct($value = null)
      *
      * If you are opening new elements, you must also close them again.
      */
-    public function xmlSerialize(Xml\Writer $writer)
+    public function xmlSerialize(Xml\Writer $writer): void
```
phil-davis, line 55: IMO this sort of change is what forces us to make a major version bump to v3 - anything that currently has `Base.php` as its parent will notice this change. `xmlSerialize` "suddenly" has a return type restricted to `void`. A child that re-implements `xmlSerialize` is likely to not be specifying any return type. PHP will start complaining when it notices that the child "specifies" a wider return type than the parent, violating LSP.

The child will have to add `: void` - so we can't release this as a 2.x version, because then composer might give the new version to projects automagically before thir code is ready.

.php-cs-fixer.dist.php
```diff
@@ -1,12 +1,13 @@
 <?php
 
-$config = PhpCsFixer\Config::create();
-$config->getFinder()
+$finder = PhpCsFixer\Finder::create()
     ->exclude('vendor')
     ->in(__DIR__);
```
staabm, line 5: ->append([
        __FILE__
    ])

lib/ContextStackTrait.php
```diff
@@ -80,28 +74,24 @@ trait ContextStackTrait
...
-     * @var array
      */
-    public $classMap = [];
+    public array $classMap = [];
 
     /**
      * Backups of previous contexts.
      *
-     * @var array
+     * @var array<int, mixed>
```
staabm, line 83: maybe `list<mixed>`

lib/Element/Base.php
staabm: Agree.

Alternatively do the PR on 2.x with all changes but these and do the BC breaking stuff separately

staabm (approved):

lib/ContextStackTrait.php
phil-davis: that works - I will push a change.

.php-cs-fixer.dist.php
phil-davis: done

lib/Element/Base.php
phil-davis: > Alternatively do the PR on 2.x with all changes but these and do the BC breaking stuff separately

IMO the other parameter type changes... are "not that exciting" and mostly just in the test code. So there is not really much value for current users that are on PHP 7.1 7.2 7.3 to get those, they don't fix anything real.

PR #223, opened 2022-08-29
Add PHP 8.2 to CI

phil-davis:
and the latest php-cs-fixer wants to change the order of some PHPdoc comments, so let it do it.


phil-davis:
https://github.com/sabre-io/xml/runs/8069173233?check_suite_focus=true
phpunit OK (111 tests, 141 assertions) with PHP  8.2 - good.

PR #224, opened 2022-09-26
Adjust new things from cs-fixer and phpstan

phil-davis:
cs-fixer now finds a bit of PHP-doc to adjust.

phpstan no longer reports some false-positives - remove them from `ignoreErrors`

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,16 +1,6 @@
 parameters:
 	ignoreErrors:
 		-
-			message: "#^Negated boolean expression is always false\\.$#"
-			count: 1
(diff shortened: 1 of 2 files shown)


phpstan-baseline.neon
```diff
@@ -1,15 +1,5 @@
 parameters:
 	ignoreErrors:
-		-
-			message: "#^Negated boolean expression is always false\\.$#"
```
staabm, line 4: do we have more errors you consider a false-positive in this file?
I can have a look into improving the phpstan side of things

phil-davis: I created issue #227 for the last item in phpstan ignoreErrrors.

PR #228, opened 2022-09-26
Prepare release 4.0.0

phil-davis:
`sabre/uri` major version 3 has a breaking change (which only applies to Windows-style URIs, but it is breaking for an installation that really uses that). So bump the major version here in `sabre/xml` so that consumers can choose exactly what combination of `sabre/xml` plus `sabre/uri` that they use.

It's a bit sad that I released 3.0.0 only a month ago, but the numbers are integers, and integers go on forever - we won't run out!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 ChangeLog
 =========
 
+4.0.0 (2022-09-26)
+------------------
+* #226 Allow sabre/uri major version 3 ( @phil-davis )
+
 3.0.0 (2022-08-17)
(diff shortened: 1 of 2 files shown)


phil-davis:
> lgtm. are there any open issues which can/should be fixed with this new major, since we are already at it?

I don't see anything that can be made ready easily. They all look like discussions from years ago that "someone" (tm) would have to dig into and understand again. And probably many would be enhancements that would be designed in a backward-compatible way anyhow (so would become 4.1 4.2 etc.

PR #229, opened 2022-10-05
Fixed error where valueobject has eaten too much

mrbig:
This fixes a bug, when valueobject is an empty element like <foo></foo> In this situation the deserializer did read one more, thus ending up a level higher than required.

I have also attached a unit test to demonstrate the problem. With the original code this xml did throw an error.

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -224,7 +224,7 @@ function valueObject(Reader $reader, string $className, string $namespace): obje
                 $reader->next();
             }
         } else {
-            if (!$reader->read()) {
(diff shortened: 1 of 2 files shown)


evert:
Pretty wild that people are still finding bugs, but LGTM!

evert (approved):

mrbig:
> Pretty wild that people are still finding bugs, but LGTM!

agree with that, I was surprised, that this haven't been found yet.

Thanks for the great software!

staabm (approved):

phil-davis:
This has been here for a while!
I will have a look and see if I can make bugfix patch releases for major versions 2, 3 and 4.

phil-davis:
Backport to v3 branch is in PR #241

PR #239, opened 2023-06-09
Cs fixer CI changes v3

phil-davis:
php-cs-fixer now wants various changes.
I have got most of this from what is in master branch (major version 4) and applied it to the major version 3 branch. See master PRs #231 #233 #234 #236 #238

### DIFF
--- a/lib/LibXMLException.php
+++ b/lib/LibXMLException.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Sabre\Xml;
 
 use LibXMLError;
-use Throwable;
 
 /**
  * This exception is thrown when the Reader runs into a parsing error.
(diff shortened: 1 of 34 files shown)


phil-davis:
@staabm this gets the major version 3 branch up-to-date with all the tools...
That will allow patch releases 3.0.* to be made out of the v3 branch as needed.
(an major version 4 patch releases come from master)

staabm (approved):

lib/LibXMLException.php
```diff
@@ -30,9 +29,9 @@ class LibXMLException extends ParseException
      *
      * You should pass a list of LibXMLError objects in its constructor.
      *
-     * @param LibXMLError[] $errors
+     * @param \LibXMLError[] $errors
```
staabm, line 32: interessting that these got prefixed

PR #242, opened 2023-06-28
[2.1] Valueobject runaway v2.1

phil-davis:
This fixes a bug, when valueobject is an empty element like <foo></foo> In this situation the deserializer did read one more, thus ending up a level higher than required

Backport PR #229 to 2.1 branch.

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -216,7 +216,7 @@ function valueObject(Reader $reader, string $className, string $namespace)
                 $reader->next();
             }
         } else {
-            if (!$reader->read()) {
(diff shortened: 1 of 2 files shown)


phil-davis:
This old branch has Travis CI (which does not run any more)
I ran the unit tests locally with PHP 7.4
```
phil@phil-Inspiron-5468:~/git/sabre-io/xml$ ./bin/phpunit --configuration tests/phpunit.xml.dist
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

...............................................................  63 / 108 ( 58%)
.............................................                   108 / 108 (100%)

Time: 348 ms, Memory: 6.00MB

OK (108 tests, 126 assertions)
```

PR #247, opened 2023-06-28
Enhance PHPdoc of Service write value parameter

### LINKED ISSUE #237: Type annotation for Service::write() is not correct
According to the [extended syntax](https://sabre.io/xml/writing/) (see for example the example after "You can even mix these syntaxes"), it is possible to pass a list (array with integer keys) as value parameter to the Service::write() method. However, the type annoation of the value parameter does not allow it:

```php
    /*
     * @param string|array<string, mixed>|object|XmlSerializable $value
     */
    public function write(string $rootElementName, $value, string $contextUri = null): string
```

This causes type error with static analysis (using psalm in my case).

phil-davis:
Fixes #237 

This makes the PHPdoc of Service write allow an array with integer keys to be passed in the `value` parameter.
That is passed to:
`$w->writeElement($rootElementName, $value);`

And Writer `writeElement` already has PHPdoc that allows `array<int|string, mixed>`

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -198,7 +198,7 @@ class Service
      * This allows an implementor to easily create URI's relative to the root
      * of the domain.
      *
-     * @param string|array<string, mixed>|object|XmlSerializable $value
+     * @param string|array<int|string, mixed>|object|XmlSerializable $value
      */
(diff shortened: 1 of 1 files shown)


staabm (approved):
lgtm

PR #251, opened 2023-06-29
Increase phpstan from level 6 to level 8

phil-davis:
This seems to be possible without breaking the code ;)

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -346,6 +346,13 @@ function functionCaller(Reader $reader, callable $func, string $namespace)
     }
 
     $funcArgs = [];
+    /**
+     * Even if $func gets "exploded" into an array, that array is still "callable"
+     * It should have elements that are the class and the function/method in the class.
+     * The declaration here helps phpstan to understand.
+     *
+     * @var (array<int,string>&callable():mixed)|\Closure|callable-string $func
+     */
     $func = is_string($func) && false !== strpos($func, '::') ? explode('::', $func) : $func;
     $ref = is_array($func) ? new \ReflectionMethod($func[0], $func[1]) : new \ReflectionFunction($func);
     foreach ($ref->getParameters() as $parameter) {
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -351,28 +351,18 @@ function functionCaller(Reader $reader, callable $func, string $namespace)
     /**
      * Even if $func gets "exploded" into an array, that array is still "callable"
      * It should have elements that are the class and the function/method in the class.
-     * The declaration here helps phpstan to understand.
-     * We check that the class/method/function exists before passing $func to create the Reflection objects.
-     *
-     * @var callable $func
      */
     $func = is_string($func) && false !== strpos($func, '::') ? explode('::', $func) : $func;
-    if (is_array($func)) {
-        if (\class_exists($func[0]) && \method_exists($func[0], $func[1])) {
+    if (is_callable($func)) {
+        if (is_array($func)) {
             $ref = new \ReflectionMethod($func[0], $func[1]);
-        } else {
-            throw new \InvalidArgumentException(__METHOD__." class '".$func[0]."' with method '".$func[1]."' does not exist.");
-        }
-    } else {
-        if (($func instanceof \Closure) || (is_string($func) && function_exists($func))) {
+        } elseif (($func instanceof \Closure) || is_string($func)) {
             $ref = new \ReflectionFunction($func);
         } else {
-            if (is_string($func)) {
-                throw new \InvalidArgumentException(__METHOD__." function '$func' does not exist.");
-            } else {
-                throw new \InvalidArgumentException(__METHOD__.' function passed in is not a valid Closure.');
-            }
+            throw new \InvalidArgumentException(__METHOD__.' func parameter is not a callable array, string or closure.');
         }
+    } else {
+        throw new \InvalidArgumentException(__METHOD__.' func parameter is not callable.');
     }
     foreach ($ref->getParameters() as $parameter) {
         $funcArgs[$parameter->getName()] = null;
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -337,7 +337,7 @@ function mixedContent(Reader $reader): array
  *
  * @return mixed whatever the 'func' callable returns
  *
- * @throws \InvalidArgumentException
+ * @throws \InvalidArgumentException|\ReflectionException
  */
 function functionCaller(Reader $reader, callable $func, string $namespace)
 {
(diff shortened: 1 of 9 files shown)


lib/Deserializer/functions.php
```diff
@@ -346,6 +346,13 @@ function functionCaller(Reader $reader, callable $func, string $namespace)
     }
 
     $funcArgs = [];
+    /**
+     * Even if $func gets "exploded" into an array, that array is still "callable"
+     * It should have elements that are the class and the function/method in the class.
+     * The declaration here helps phpstan to understand.
+     *
+     * @var (array<int,string>&callable():mixed)|\Closure|callable-string $func
```
phil-davis, line 354: without telling phpstan all about $func it thinks that the array returned by `explode()` is "any old array".
And phpstan also doesn't think that `ReflectionFunction` is being called with a valid thing.

lib/Element/Base.php
```diff
@@ -73,7 +73,7 @@ public function xmlSerialize(Xml\Writer $writer): void
      * $reader->parseInnerTree() will parse the entire sub-tree, and advance to
      * the next element.
      *
-     * @return array<string, mixed>|string|null
+     * @return array<int,array<string, mixed>>|string|null
```
phil-davis, line 76: This was an existing error in the PHPdoc. This thing returns an array of arrays.

lib/Reader.php
```diff
@@ -105,7 +105,7 @@ public function parse(): array
      *
      * @param array<string, mixed>|null $elementMap
      *
-     * @return array<string, mixed>
+     * @return array<int,array<string, mixed>>
```
phil-davis, line 108: This was an existing error in the PHPdoc. This thing returns an array of arrays.

lib/Reader.php
```diff
@@ -130,7 +130,7 @@ public function parseGetElements(array $elementMap = null): array
      *
      * @param array<string, mixed>|null $elementMap
      *
-     * @return array<string, mixed>|string|null
+     * @return array<int,array<string, mixed>>|string|null
```
phil-davis, line 133: This was an existing error in the PHPdoc. This thing returns an array of arrays.

lib/Element/Uri.php
```diff
@@ -59,7 +59,7 @@ public function xmlSerialize(Xml\Writer $writer): void
     {
         $writer->text(
             resolve(
-                $writer->contextUri,
+                $writer->contextUri ?? '',
```
phil-davis, line 62: `contextUri` starts life set to `null`.
phpstan can't be sure that it is no longer `null`
The safest thing here is to pass the empty string if it is currently `null`

lib/Reader.php
```diff
@@ -298,14 +298,14 @@ public function getDeserializerForElementName(string $name): callable
...
         $deserializer = $this->elementMap[$name];
-        if (is_subclass_of($deserializer, 'Sabre\\Xml\\XmlDeserializable')) {
-            return [$deserializer, 'xmlDeserialize'];
-        }
-
         if (is_callable($deserializer)) {
             return $deserializer;
         }
 
+        if (is_subclass_of($deserializer, 'Sabre\\Xml\\XmlDeserializable')) {
```
phil-davis, line 305: `is_subclass_of` takes a class name or object in parameter 1.
But `$deserializer` can be a callable function.
phpstan is happy if the check for `is_callable` comes first.

lib/Service.php
```diff
@@ -241,6 +241,7 @@ public function write(string $rootElementName, $value, string $contextUri = null
     public function mapValueObject(string $elementName, string $className): void
     {
         list($namespace) = self::parseClarkNotation($elementName);
+        $namespace = $namespace ?? '';
```
phil-davis, line 244: `parseClarkNotation` says that it can return `string|null` in the first element of the returned array.
later uses of `$namespace` expect a `string`. So just convert `null` to the empty string here.

Same for a few place in Writer.php below.

tests/Sabre/Xml/ServiceTest.php
phil-davis, line 1: Adding a few asserts in the tests helps phpstan to be confident that the vars are of the correct types.
And it will make the tests fail early if something unexpected happens.

lib/Deserializer/functions.php
phil-davis: I added a bunch of checks in the latest commit, and adjusted the PHPdoc var declaration to just say that it is a callable.

phil-davis:
I will also try to add some unit tests that pass non-existent functions/methods to `functionCaller` and see if I can cover the exceptions that should be thrown.

lib/Deserializer/functions.php
```diff
@@ -346,8 +348,32 @@ function functionCaller(Reader $reader, callable $func, string $namespace)
...
+     * It should have elements that are the class and the function/method in the class.
+     * The declaration here helps phpstan to understand.
+     * We check that the class/method/function exists before passing $func to create the Reflection objects.
+     *
+     * @var callable $func
+     */
     $func = is_string($func) && false !== strpos($func, '::') ? explode('::', $func) : $func;
-    $ref = is_array($func) ? new \ReflectionMethod($func[0], $func[1]) : new \ReflectionFunction($func);
+    if (is_array($func)) {
+        if (\class_exists($func[0]) && \method_exists($func[0], $func[1])) {
```
staabm, line 361: after seeing this big block of code, I got the idea that we maybe can just use `is_callable()`?

phil-davis: I rewrote the block so it starts with "if is_callable".
Maybe that makes it a bit easier to read?
And it means that there does not need to be a PHPdoc of `$func` declaring that it is `callable`

I still need to check the detail of what sort of callable it is, to choose whether to call `\ReflectionMethod` or `\ReflectionFunction`

staabm: looks good, thanks

lib/Deserializer/functions.php
```diff
@@ -346,8 +348,22 @@ function functionCaller(Reader $reader, callable $func, string $namespace)
...
+    if (is_callable($func)) {
+        if (is_array($func)) {
+            $ref = new \ReflectionMethod($func[0], $func[1]);
+        } elseif (($func instanceof \Closure) || is_string($func)) {
+            $ref = new \ReflectionFunction($func);
+        } else {
+            throw new \InvalidArgumentException(__METHOD__.' func parameter is not a callable array, string or closure.');
+        }
+    } else {
+        throw new \InvalidArgumentException(__METHOD__.' func parameter is not callable.');
```
phil-davis, line 365: I can't see a way to get to this code.
When I try to make a unit test that passes a function reference like "SomeValidClass::unknownMethod", then PHP notices that during the call to `functionCaller`. Even though the string passed looks like a reasonable format for a callable thing, at run-time PHP checks if the method exists in the class. And the test fails with:
```
TypeError: Argument 2 passed to Sabre\Xml\Deserializer\functionCaller() must be callable, string given, called in /home/phil/git/sabre-io/xml/tests/Sabre/Xml/Deserializer/FunctionCallerTest.php on line 154
```

So PHP really verifies that the thing passed in as parameter `$func` not only looks like it could be callable, but that it is actually callable.

So in real-life we don't need all this double-checking.

I will see if there is a simpler way of coding this and keeping phpstan happy at the same time.

lib/Deserializer/functions.php
```diff
@@ -346,8 +348,20 @@ function functionCaller(Reader $reader, callable $func, string $namespace)
...
+    } elseif (is_string($func) && false !== strpos($func, '::')) {
+        // We have a string that should refer to a method that exists, like "MyClass::someMethod"
+        // ReflectionMethod knows how to handle that as-is
+        $ref = new \ReflectionMethod($func);
+    } elseif ($func instanceof \Closure || is_string($func)) {
+        // We have an actual Closure (a real function) or a string that is the name of a function
+        // ReflectionFunction can take either of those
+        $ref = new \ReflectionFunction($func);
+    } else {
+        throw new \InvalidArgumentException(__METHOD__.' unable to use func parameter with ReflectionMethod or ReflectionFunction.');
```
phil-davis, line 362: I can't cover this line with a unit test.
My analysis is:
PHP always checks that `$func` is an actual callable that will work (it doesn't just check that `$func` has a format that looks like it should be callable, does actual check that something like "MyClass::someMethod" really is a function that exists in a class that exists.

So when we get to this "if" block, we know that `$func` is a working callable.

If it is in array form, then we call `ReflectionMethod`
If it is in string form with a "::" in it, then we also call `ReflectionMethod`
If it is a real Closure, or a string without "::" in it then we call `ReflectionFunction` (because it is an ordinary function)

And those are all the ways that a "callable" can be passed.

But if I do not have an "else" to catch Unicorns, then `phpstan` complains that `$ref` might not be defined.

I think that it is a computable task for `phpstan` to analyze that `$ref` must be defined. `phpstan` has to understand what are all the things whose union is a "callable". Then it can deduce that the if-elseif-... covers them all.

phil-davis:
@staabm I reorganized the if-else... block in `functionCaller` so that it does not need to explode strings like "MyClass::someMethod". That avoids phpstan getting confused about what the exploded array is (it is still callable, but phpstan can't guess that).

I thin kthis is enough on this! Please review and I will merge if OK.

staabm (approved):
thank you

PR #270, opened 2024-02-08
Fixed handling of elements from foreign namespaces in values object (2.2)

mrbig:
When encountering an element from another namespace the value object parser does skip the opening element. However when it encounters the closing element it did handle it like it was the closing element of the one being processed.

This commit and unit test fixes the issue by ignoring the element completely. Without the patch the unit test would not parse the `lastName` element.

This patch is against the branch `2.2`, because I'm still using this one, I hope it's not a problem. I've checked the `master` branch as well, and it should be straightforward to apply it there as well.

BTW: would it be possible to add mappings for multiple namespaces?

### DIFF
--- a/lib/Deserializer/functions.php
+++ b/lib/Deserializer/functions.php
@@ -215,6 +215,9 @@ function valueObject(Reader $reader, string $className, string $namespace)
                 // Ignore property
                 $reader->next();
             }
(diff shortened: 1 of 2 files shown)


phil-davis:
Forward-port to master is PR #273

staabm (approved):

PR #271, opened 2024-02-11
chore: use php-cs-fixer 3.49

phil-davis:
php-cs-fixer 3.49 (or some recent minor release) wants to use the `?` (nullable) syntax on optional parameters.
That syntax has been supported for quite a while since PHP https://www.php.net/manual/en/migration71.new-features.php

In the cases here, as well as the default value of the parameter being `null`, the `?` explicitly allows the caller to pass the value `null` if they want.

This seems reasonable, I don't see how it can break any existing usage.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -44,7 +44,7 @@
         }
     },
     "require-dev": {
-        "friendsofphp/php-cs-fixer": "^3.49",
+        "friendsofphp/php-cs-fixer": "^3.51",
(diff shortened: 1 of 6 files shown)


phil-davis:
Has a minor complaint about project code coverage - strange report, the codecov/patch reports 100% of diff hit.

phil-davis:
Bumped to use cs-fixer 3.51

PR #272, opened 2024-04-16
More precise `parseClarkNotation` return type

staabm:
while updating from a old version I got the impression that the first offset of `parseClarkNotation` cannot get null.

https://3v4l.org/Bo8q3

did I miss something?

### DIFF
--- a/lib/Service.php
+++ b/lib/Service.php
@@ -241,7 +241,6 @@ class Service
     public function mapValueObject(string $elementName, string $className): void
     {
         list($namespace) = self::parseClarkNotation($elementName);
-        $namespace = $namespace ?? '';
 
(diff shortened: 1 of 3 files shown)


staabm:
sorry for trial and error commits.. working from a new laptop without proper project setup

phil-davis (approved):
correct - `parseClarkNotation` either throws an exception, or finds a match of the regex in the string, which will always return non-null entries from `$matches` (which might have the empty string as a value, like in your new unit test example)

I think that I can squash these commits and force-push before I merge this...

staabm:
> I think that I can squash these commits and force-push before I merge this...

absolutely thanks

PR #282, opened 2024-05-10
Stop exporting php-cs-fixer config

szepeviktor:
This is a development file.


phil-davis (approved):
Thanks, this got missed when php-cs-fixer was bumped to a newer version and the file name changed.

szepeviktor:
This could be checked in a CI step.
https://github.com/szepeviktor/byte-level-care/blob/6aaffb792e2f1e5f535f7bff83a969353d1966cb/.github/workflows/reusable-integrity.yml#L186-L187

PR #290, opened 2024-07-29
chore: add cs-fixer v3

phil-davis:
Add the ability to use php-cs-fixer major version 3.
Add `.php-cs-fixer.dist.php` (the settings of php-cs-fixer for major version 3) and put the `nullable_type_declaration` check into it.
Run cs-fixer on PHP 8.4.
Leave the old cs-fixer still running on PHP 7.1.
And run cs-fixer on every PHP version, to be sure. In CI, `composer` will choose whether to use major version 2 or 3 of php-cs-fixer.

The newer cs-fixer v3 makes various nice fixes that do not break on PHP 7.1.
The old cs-fixer v2 is happy with the changes, it does not try to change them back - that is nice.

This is only needed for this old release series that still has PHP 7.1 support. We can run both cs-fixer and phpstan (and phpunit) on PHP 7.1 through 8.4. I think that helps give confidence that the code works across that range of PHP versions.

Note: when I forward-port all this stuff to later major releases, I won't need to do all this tweaking because we move to supporting only PHP 7.4 and up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         php-versions: ['7.2', '7.3', '7.4', '8.0', '8.1', '8.2', '8.3']
         coverage: ['pcov']
-        code-style: ['no']
+        code-style: ['yes']
         code-analysis: ['no']
         include:
           - php-versions: '7.1'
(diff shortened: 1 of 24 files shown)


phil-davis:
I have confirmed locally with PHP 8.4 that if a remove the `?` from a parameter declaration like 
`?array $elementMap = null`
then run cs-fixer,
it finds the problem and puts the `?` back - good.

PR #296, opened 2024-09-05
Add PHP 8.4 compliance for v4

phil-davis:
Apply the small code changes and the CI and tooling changes for PHP 8.4 to master (which is where the v4 release series comes from).

And fixup `parse()` and `expect()` functions in `Service.php` - see the comments below about what `phpstan` noticed.
Now, if someone passes a closed resource, we will throw the usual `ParseException` that is thrown when someone passes an empty input.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,10 +14,16 @@ jobs:
       matrix:
         php-versions: ['8.0', '8.1', '8.2', '8.3']
         coverage: ['pcov']
+        code-style: ['no']
         code-analysis: ['no']
         include:
           - php-versions: '7.4'
             coverage: 'none'
+            code-style: 'yes'
+            code-analysis: 'yes'
+          - php-versions: '8.4'
+            coverage: 'pcov'
+            code-style: 'no'
             code-analysis: 'yes'
     steps:
       - name: Checkout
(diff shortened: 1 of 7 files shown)


phil-davis:
```
 ------ ----------------------------------------------------------------- 
  Line   lib/Service.php                                                  
 ------ ----------------------------------------------------------------- 
  125    Parameter #1 $source of method XMLReader::XML() expects string,  
         resource|string given.                                           
  169    Parameter #1 $source of method XMLReader::XML() expects string,  
         resource|string given.                                           
 ------ ----------------------------------------------------------------- 
```

Hmmm - `phpstan` is complaining about this. It did not complain in the older 2.2 and v3 branches. I will check what the code looks like there.

phil-davis:
I raised an issue with phpstan https://github.com/phpstan/phpstan/issues/11634
See the comments there. Actually the problem is in our code.

https://www.php.net/manual/en/function.is-resource.php
------------------------------
Note:
    is_resource() is not a strict type-checking method: it will return [false](https://www.php.net/manual/en/reserved.constants.php#constant.false) if value is a resource variable that has been closed.
------------------------------

When we use `is_resource()` that actually does "is open resource". The variable could still be a resource, but closed.

https://www.php.net/manual/en/function.stream-get-contents.php
If we call `stream_get_contents` with a closed resource, the worst that happens is that it returns `false` and when we cast `false` to string, that gives the empty string. And there is already exception handling for that.

So it is better, and we can safely, use `if (!is_string($input))`

staabm (approved):

PR #299, opened 2024-09-05
Check CI for 2.2

phil-davis:


phil-davis:
PHP CS Fixer 3.64.0 wants to add a trailing comma to the parameters of the `list()` "function"

But the old PHP CS fixer v2 does not require that, and in fact it tries to refactor the code removing the trailing comma.
It is getting more difficult to keep supporting all the way from PHP 7.1 to 8.4 and with the different major versions of the code style and analysis tools.

staabm:
IMO dropping 7.x (or everything below 7.4h should be considered

phil-davis:
> IMO dropping 7.x (or everything below 7.4h should be considered

Agree - I will set all the repos to a minimum of 7.4 supported. The analysis tools are all still supporting that OK, and 7.4 has a good range of the "new" code styling options available.

phil-davis:
I got it passing in #298 by pinning php-cs-fixer to 3.63.2 - that will do for now.

PR #305, opened 2024-12-02
chore(deps): phpstan major version 2

phil-davis:


lib/Serializer/functions.php
```diff
@@ -147,7 +147,7 @@ function repeatingElements(Writer $writer, array $items, string $childElementNam
  *
  * You can even mix the two array syntaxes.
  *
- * @param string|int|float|bool|array<int|string, mixed>|object $value
+ * @param string|int|float|bool|array<int|string, mixed>|object|null $value
```
phil-davis, line 150: This function actually does let you pass `null` to it, and will just "do nothing" to the Writer object. In that case there is nothing to serialize but it does not complain about that.

Telling PHPstan that `null` is possible, stops it complaining about `elseif (is_object($value))` below - it now understands that `$value` could be `null`or an object at that point.

lib/Serializer/functions.php
```diff
@@ -181,7 +181,7 @@ function standardSerializer(Writer $writer, $value): void
                 // This item has a numeric index. We just loop through the
                 // array and throw it back in the writer.
                 standardSerializer($writer, $item);
-            } elseif (is_string($name) && is_array($item) && isset($item['attributes'])) {
+            } elseif (is_array($item) && isset($item['attributes'])) {
```
phil-davis, line 184: Array keys (name) have to be `int` or `string`. And we tested for `is_int` above. So PHPstan correctly reports that the `is_string` check here is useless.

phil-davis: But as a result of looking into this, I don't understand why PHPstan allows the code at line 193, `elseif (is_string($name))`, because we know that `$name` must be a string at that point.
And the `else` at line 198 looks useless. The exception can never be thrown, because PHP only allows array keys of type `int` or `string`, so we can never get to that `InvalidArgumentException`

phpstan-baseline.neon
```diff
@@ -1,2 +1,6 @@
 parameters:
-	ignoreErrors:
+  ignoreErrors:
+  -
+    message: "#^Call to function is_null\\(\\) with null will always evaluate to true\\.$#"
+    path: lib/Serializer/functions.php
+    count: 1
```
phil-davis, line 6: This message is emitted because of line 204
```
    } elseif (!is_null($value)) {
        throw new \InvalidArgumentException('The writer cannot serialize values of type: '.gettype($value));
    }
```
The if-else cases already processed all the possible types according to the type declaration of `$value` except for `null`. So PHPstan thinks that `$value` must be `null` at that point, and so we can never get that `InvalidArgumentException`

But someone could pass some other unusual type of value to `standardSerializer` The type would not fit with the PHPdoc declared set of types, but the PHP run-time is not going to complain. So "catch-all" exceptions like that are actually valid/useful.

tests/Sabre/Xml/ServiceTest.php
```diff
@@ -182,9 +182,6 @@ public function testEmptyPropfind(string $xml): void
         $util->namespaceMap = [
             'http://sabre.io/ns' => 'stdClass',
         ];
-        /**
-         * @var PropFindTestAsset
-         */
```
phil-davis, line 187: It's better not to tell the test code what `$result` should be.
The test code does asserts itself to check that.

lib/Serializer/functions.php
phil-davis: What is the "good thing" to do for situations like this?
An input parameter allows some "complicated" set of inputs/types.
It is too complicated to be expressed in the actual PHP function declaration.
So at run-time PHP will actually let the caller pass anything in the parameter.
The function code checks the input parameter, doing "the expected thing" to process inputs of each of the "officially allowed" inputs.

The function code throws an exception if the passed-in value is not one of the "officially allowed" inputs. But PHPstan is going to consider that that is "useless" code.

staabm: For this defensive style of programming you should set `treatPhpDocTypesAsCertain: false` in the PHPStan config

phil-davis: If I do that, it will apply to all the code. So I just added the entry to `ignoreErrors`

lib/Serializer/functions.php
staabm: array keys are pretty magic and some auto-conversion are going on in the background (some strings are turned into numbers)

https://3v4l.org/amqMg

lib/Serializer/functions.php
staabm: sure, if you are fine with ignoring the error thats fine, too.

lib/Serializer/functions.php
phil-davis: > array keys are pretty magic and some auto-conversions are going on in the background in php-src (some strings are turned into numbers)

If someone tries to have a key `$someArray['42']` then PHP is going to treat it as `$someArray[42]`, and the code (both before and after this PR) is going to go into the `is_int()` case a few lines up. So that's a "feature" of PHP. And I don't think that there will be any callers who actually want/need to force a string key that is an ASCII decimal number.

PR #324, opened 2026-04-27
chore: drop PHP less than 8.2

phil-davis:
I think I have gone too far with this.

`phpstan` tells me that the XML method of XMLReader is statis nowadays, and that calls should be like:
```
$reader = Reader::XML($input);
```

But a call like that returns an XMLReader object. We want a Sabre\Xml\Reader object that has extra methods available in it ( getClark() parse() etc.).

So I have tried overriding `XMLReader::XML` with an implementation in `Reader::XML`. That calls `parent::XML` and then makes an attempt to copy all the object vars from the returned `XMLReader` object to a new `Reader` object and reutrn that.

I have got cs-fixer, rector, and phpstan passing.

But `phpunit` complains:
`Error: Data must be loaded before reading`

So I haven't quite got the `Reader` object to be set up correctly.

I tried other ways to do this with Reflection, but no joy in copying the attributes. Maybe there is some other way to do this, or maybe this is entirely the wrong approach!

I will now make a less ambitious PR, and someone might give me clues about what approach should happen here.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,14 +12,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.0', '8.1', '8.2', '8.3','8.4']
+        php-versions: ['8.3','8.4']
         coverage: ['pcov']
         code-style: ['no']
         code-analysis: ['no']
         rector-check: ['no']
         include:
-          - php-versions: '7.4'
-            coverage: 'none'
+          - php-versions: '8.2'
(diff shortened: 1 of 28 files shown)


phil-davis:
Adding `phpstan/phpstan-strict-rules` caused quite a few good messages to be fixed/adjusted.
But also reported a lot of:
`Dynamic call to static method XMLReader::XML()`

and "fixing" that is what is causing me pain here.

phil-davis:
see #325

PR #325, opened 2026-04-27
chore: support PHP 8.2 and up

phil-davis:
I have done everything from #324 but ignored:
`Dynamic call to static method XMLReader::XML`

That seems to allow for lots of tidy-up of bits of code, keeping unit tests passing, without getting down the rabbit hole of whether to change lots of `$reader->XML($input)` to somehow be like:
```
$reader = Reader::XML($input);
```

`phpstan-strict-rules` is enabled, so that seems "a good thing".
`rector` made a lot of these changes itself.

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -3,7 +3,7 @@ includes:
 
 parameters:
   level: 8
-  phpVersion: 70430 # PHP 7.4.30
+  phpVersion: 80200 # PHP 8.2.0
   paths:
     - lib
     - tests
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -7,3 +7,7 @@ parameters:
   paths:
     - lib
     - tests
+  ignoreErrors:
+    -
+      message: '#^Dynamic call to static method XMLReader::XML\(\)\.$#'
+      identifier: staticMethod.dynamicCall
--- a/tests/Sabre/Xml/Deserializer/ValueObjectTest.php
+++ b/tests/Sabre/Xml/Deserializer/ValueObjectTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sabre\Xml\Deserializer;
 
+use PHPUnit\Framework\Assert;
 use PHPUnit\Framework\TestCase;
 use Sabre\Xml\Reader;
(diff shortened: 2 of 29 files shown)


phpstan.neon
```diff
@@ -3,7 +3,11 @@ includes:
...
   level: 8
-  phpVersion: 70430 # PHP 7.4.30
+  phpVersion: 80200 # PHP 8.2.0
   paths:
     - lib
     - tests
+  ignoreErrors:
+    -
+      message: '#^Dynamic call to static method XMLReader::XML\(\)\.$#'
+      identifier: staticMethod.dynamicCall
```
staabm, line 13: please reproduce on phpstan.org/try

tests/Sabre/Xml/Deserializer/ValueObjectTest.php
```diff
@@ -207,7 +208,7 @@ public function testDeserializeValueObjectEmptyString(): void
             'attributes' => [],
         ];
 
-        $this->assertEquals(
+        Assert::assertEquals(
```
staabm, line 211: I also prefer static assert methods, because that way PHPStan can see that this method-calls will not change/invalidate types gathered in scope.

this can make improve phpstan results (or ease handling)

:+1:

staabm (approved):
nice

phpstan.neon
phil-davis: @staabm phpstan is correct. XMLReader::XML is static in PHP8.
So it should really be called statically.
But it does still work when called dynamically, and does update the "internals" of the object that it is called on when called dynamically. But that is not that clearly documented:
https://www.php.net/manual/en/xmlreader.xml.php
The Changelog has an entry:
8.0.0 	XMLReader::XML() is now declared as static method, but can still be called on an [XMLReader](https://www.php.net/manual/en/class.xmlreader.php) instance.

but there is no information about how long this "old" behavior will be supported. I suppose for "a long time", since there is no deprecation announced.

So I'm not sure what to "complain" about to phpstan.

staabm: ok - fine than. thanks for the explanation.

PR #328, opened 2026-05-31
test: fix phunit10 deprecations

phil-davis:
phpunit10 was reporting deprecations. It wants dataProvider methods to be declared static, so that is fixed here.

Also, converted phpunit.xml to version 10 format.
And set options failOnRisky and failOnWarning so that CI will fail in these cases. (That is new phpunit behavior that needs to be set, otherwise the test suite can exit with success even though there are "risky" tests or warnings)

### DIFF
--- a/tests/Sabre/Xml/Element/XmlFragmentTest.php
+++ b/tests/Sabre/Xml/Element/XmlFragmentTest.php
@@ -56,7 +56,7 @@ BLA;
      *
      * @return array<int, array<int, string>>
      */
-    public function xmlProvider(): array
+    public static function xmlProvider(): array
     {
         return [
             [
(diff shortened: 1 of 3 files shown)


phil-davis:
Note: test code only, nothing to release.
phpunit output:
```
PHPUnit 10.5.63 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.31 with PCOV 1.0.12
Configuration: /home/runner/work/xml/xml/tests/phpunit.xml

...............................................................  63 / 116 ( 54%)
.....................................................           116 / 116 (100%)

Time: 00:00.132, Memory: 16.00 MB

OK (116 tests, 150 assertions)
```

No deprecations or warnings.

PR #336, opened 2026-07-07
ci: add PHP 8.6

phil-davis:
leaving the other test matrices for PHP unchanged for now.

Run all CI checks on PHP 8.6.

This is just a draft PR to check that CI passes. When PHP 8.6 gets to the RC stage, then I will make a real PR to add 8.6.


.github/workflows/ci.yml
```diff
@@ -28,6 +28,11 @@ jobs:
             code-style: 'no'
             code-analysis: 'yes'
             rector-check: 'yes'
+          - php-versions: '8.6'
+            coverage: 'pcov'
+            code-style: 'no'
+            code-analysis: 'yes'
+            rector-check: 'yes'
```
staabm, line 35: I would be cautios with rector.. its likely bringing up some problems until it properly supports 8.6 which is still alpha

PR #338, opened 2026-07-15
feat: Allow to stream input

### LINKED ISSUE #337: Make use of XMLReader::fromStream
Available since php 8.4

CarlSchwan:
Fixes #337


phpstan-rules-lt-8.4.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+  ignoreErrors:
+  -
+    message: "#^Call to an undefined static method Sabre\Xml\Reader::from#"
```
phil-davis:
```suggestion
    message: "#^Call to an undefined static method Sabre\\Xml\\Reader::from#"
```

I think that the literal `\` need to be escaped `\\`

phpstan.conditional.php
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * Add conditionally extra phpstan rules
+ * Typically used for specific PHP versions
+ * Must be called in main PHPStan neon config file.
+ */
+$includes = [];
+
+// For versions smaller that PHP 8.4.
+if (PHP_VERSION_ID < 80300) {
```
phil-davis:
```suggestion
if (PHP_VERSION_ID < 80400) {
```

phpstan-rules-lt-8.4.neon
staabm: you can use `rawMessage` (which 1:1 matches strings) instead of `message` so you don't need regex patterns which is faster and does not require escaping

phil-davis (dismissed):
LGTM

staabm (commented):
it seems most of the changes miss test-coverage?

phpstan.conditional.php
```diff
@@ -0,0 +1,21 @@
...
+declare(strict_types=1);
+
+/**
+ * Add conditionally extra phpstan rules
+ * Typically used for specific PHP versions
+ * Must be called in main PHPStan neon config file.
+ */
+$includes = [];
+
+// For versions smaller that PHP 8.4.
```
staabm: useless comment

```suggestion
```

tests/Sabre/Xml/ServiceTest.php
```diff
@@ -364,9 +364,11 @@ public static function provideParseClarkNotationInput(): iterable
      */
     public static function providesEmptyInput(): array
     {
-        $emptyResource = fopen('php://input', 'r');
         $data = [];
-        $data[] = [$emptyResource];
+        if (PHP_VERSION_ID < 80400) {
```
staabm, line 383: this test was changed in a way, that a feature which worked on "all supported php versions" before now only is tested on PHP < 8.4?

do we need assertions for the same test on PHP_VERSION_ID >= 80400?

CarlSchwan: Yeah, now we can't check if the input stream is empty because that would require reading the whole stream which we can't 

Maybe we could then throw an error after the parsing if the result is empty?

CarlSchwan: Changed a bit the tests. Now the test case of empty stream will catch the LibXMLException (which inherit from the ParseException) and the LibXMLException message is correctly formatted

